PR #25, opened 2015-02-05
add Double/Text typed ranges

unthingable:


bitemyapp:
This looks promising.

Is there any use for a textual range that isn't a datefield?

unthingable:
Strings and integers. Possibly more.

bitemyapp:
Integers are covered by Double, since it's all JSON number anyway.

What does a range on a String mean? What are they expecting?

unthingable:
_Matches documents with fields that have terms within a certain range. The type of the Lucene query depends on the field type, for string fields, the TermRangeQuery, while for number/date fields, the query is a NumericRangeQuery._

IMHO, since the user already has to be aware of the field type in a range query (which Bloodhound doesn't know), it's not a stretch for them to know what they're expecting from a query. I don't believe it needs to be any more explicitly typed.

I have code that depends on this change and I'm quite happy with it.

bitemyapp:
@unthingable the type-safety enforced by Bloodhound isn't very complicated or fancy, but it *is* valuable and a PR that unnecessarily compromises that is not something I'm apt to merge.

Based on Elasticsearch's documentation and what I can find on Stack Overflow, ranges are only well defined for number and date fields. That means the two types for Ranges should be a date type (possibly with TZ given the textual representations I am seeing) or a number type. Not Text.

The library should handle the correctness of the serialization of a date value into the textual format (Looks like ISO?) that Elasticsearch expects.

unthingable:
As a user of the library I can assure you that I _like_ it as Text, especially when I'm testing things. I can query on a date field like `RangeGte "2014"` and it will interpret it as as "2014-01-01 00:00" for me. Do I want to have to create a piece of data of a proper date type from a string "2014"? No, I do not. I'm really happy with `RangeGte "2014"` and being able to do that is valuable to me. For me, text is an fine datatype, in this case. 

Not that I would mind having a proper Date range with TZ support, but I would also want to keep the plain text range, for me to decide when to use. Can't we have that?

tonymorris:
As a user and provider of libraries, I can assure you right back that you don't want text, especially when you are "testing things." This illusion of convenience does not exist, precisely because there is already excellent library support for viewing a data type proper "as if it were a string." We call these things (libraries) lenses and prisms and so on.

In this case, since safety is paramount, and although we would not sacrifice it for this convenience anyway, we *don't need to*, and we don't need to *precisely because of library support*.

mwotton:
strong -1 from me as another user of the library. Calendar & date bugs are some of the worst you hit, I need the strongest protection I can get.

unthingable:
Very well, point taken. @bitemyapp, apologies for having inadvertently vented some frustration, and for trying to rush an unsafe feature.

I may lack the expertise to do this properly (don't know how to range over dates as numbers, for example), but I could see a RangeDate type, serializing to strings. What is the preferred type for the dates?

Is the consensus then that a text range has no right to exists, or could it be renamed to UnsafeRangeText and spared?

tonymorris:
@unthingable If we step back and ask what the goal is, we might talk about achieving the conveniences of your proposal (or better, surpassing them), as well as the safety and documentation that types provide.

Is it possible to do this? Yes, it is, but the execution is radically different to "concede the point and use strings."

Instead, it's the exploitation of diverse library support. So you might ask how to recognise the specifics of what library support is appropriate here. I can only answer with a humble, "by practising." That is to say, learn about all the different *fundamental* libraries and tools (differentiating fundamental from junk is initially hard, I admit), then apply such that when you are faced with a practical problem, you are equipped to determine which of these tools to deploy.

In the case here, and without knowing the specifics of the bloodhound library itself, the problem description strongly suggests something like a prism (`Control.Lens.Prism` to a string). Learning what this is, how to use it and how it achieves all of our initial goals is a personal goal to potentially aspire to. I highly recommend it.

unthingable:
@bitemyapp Something like that?

bitemyapp:
@unthingable late-night initial impression: seems promising. Tests are broken and some cosmetic improvement seems possible. I'll review during the day and say more then.

Thank you!

unthingable:
Oops, forgot to rerun the tests one last time.

bitemyapp:
@unthingable while you're trying to figure out the tests passing, a design change would be to clean up the very big flat sum into a mixture of sums and products like it used to be.

I'm still thinking about the best way to do this.

bitemyapp:
Having explored the GADT/ExistentialQuantification solution, having a flat sum type is better for now.

unthingable:
Excellent. Thank you!

bitemyapp:
@unthingable tests are passing. Remaining issues:

The constructor arguments aren't as type-safe as they were before. That's why there were separate LessThan/GreaterThan/LTE/GTE newtypes. 

These transposition errors will be exacerbated given constructor names like `RangeDateGteLte`, it's lower then upper in the arguments, but upper then lower bound in the name of the constructor.

There's no test, AFAICT, for the UTCTime ranges. I also can't see how UTCTime is getting serialized for the queries in the code.

This is a good start, but it's not mergeable yet.

unthingable:
So it works for seconds. After that UTCTime jumps straight into picoseconds, which ES does not handle.

unthingable:
The newtypes have been restored, dropping them was unwise.

bitemyapp:
@unthingable I'm going to merge this, make some minor tweaks, and look into millisecond support. When that's all done, I'll push a release and notify you when the release is on Hackage.

PR #27, opened 2015-03-04
26 bhenv

MichaelXavier:
I can flatten this and whatnot, this is mainly for review.

### DIFF
--- a/tests/tests.hs
+++ b/tests/tests.hs
@@ -6,6 +6,7 @@ module Main where
 
 import           Control.Applicative
 import           Control.Monad
+import           Control.Monad.Reader
 import           Data.Aeson
 import qualified  Data.HashMap.Strict      as HM
 import           Data.List                 (nub)
@@ -34,15 +35,23 @@ testIndex   = IndexName "bloodhound-tests-twitter-1"
 testMapping :: MappingName
 testMapping = MappingName "tweet"
 
+
+--TODO: similar convenience function in Client
+-- withTestEnv :: MonadIO m => Reader m a -> IO a
+withTestEnv f = withManager defaultManagerSettings $ \mgr -> do
+  let env = BHEnv { bhServer = testServer
+                  , bhManager = mgr }
+  runReaderT f env
+
 validateStatus :: Response body -> Int -> Expectation
 validateStatus resp expected =
   (NHTS.statusCode $ responseStatus resp)
   `shouldBe` (expected :: Int)
 
-createExampleIndex :: IO Reply
-createExampleIndex = createIndex testServer defaultIndexSettings testIndex
-deleteExampleIndex :: IO Reply
-deleteExampleIndex = deleteIndex testServer testIndex
+-- createExampleIndex :: IO Reply
+createExampleIndex = createIndex defaultIndexSettings testIndex
+-- deleteExampleIndex :: IO Reply
+deleteExampleIndex = deleteIndex testIndex
 
 data ServerVersion = ServerVersion Int Int Int deriving (Show, Eq, Ord)
 
@@ -70,18 +79,16 @@ mkServerVersion [majorVer, minorVer, patchVer] =
   Just (ServerVersion majorVer minorVer patchVer)
 mkServerVersion _                 = Nothing
 
-getServerVersion :: Server -> IO (Maybe ServerVersion)
-getServerVersion s = liftM extractVersion (getStatus s)
+-- getServerVersion :: IO (Maybe ServerVersion)
+getServerVersion = liftM extractVersion (withTestEnv getStatus)
   where
     version'                    = T.splitOn "." . number . version
     toInt                       = read . T.unpack
     parseVersion v              = map toInt (version' v)
     extractVersion              = join . liftM (mkServerVersion . parseVersion)
 
-
-
 testServerBranch :: IO (Maybe ServerVersion)
-testServerBranch = getServerVersion testServer >>= \v -> return $ liftM serverBranch v
+testServerBranch = getServerVersion >>= \v -> return $ liftM serverBranch v
 
 atleast :: ServerVersion -> IO Bool
 atleast v = testServerBranch >>= \x -> return $ x >= Just (serverBranch v)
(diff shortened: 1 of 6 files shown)


tests/tests.hs
```diff
@@ -34,15 +35,18 @@ testIndex   = IndexName "bloodhound-tests-twitter-1"
 testMapping :: MappingName
 testMapping = MappingName "tweet"
 
+withTestEnv :: BH a -> IO a
+withTestEnv = withBH defaultManagerSettings testServer
```
bitemyapp: This is really good. Low hassle but still abstracts things out.

bitemyapp:
This looks good, my only concern (I know this is painful, I'm sorry) will be documentation so people can follow along with examples and have them work.

I started, but haven't finished, porting the README examples into doctests that get executed. You probably saw the beginning of that in the Client module.

You have any thoughts on the doc situation?

Are you happy with this design?

MichaelXavier:
As far as the code design goes, @ozataman is going to take a look when he gets a chance. For myself, I think it was easy to use and understand. This should remove a lot of argument noise from the end user's code. Further, if we need to add more fields to the environment at some point, this design seems superior. Without this design, the end user will have to change every call site mechanically. It is also really nice to be able to get pipelining for free and embed that in another monad stack. This does up the barrier for entry slightly because the user will probably have to know about monad transformers but I think there's little overlap between people using elasticsearch from haskell and ones who have never seen a ReaderT before.

As far as documentation, I'm a fan of doctest when it actually works, and you seem to have figured out the dependency issues. Are you saying you're going to have doctest actually execute your readme in addition to your source file doctest?

ozataman:
Nice, it's looking good. Here are a few suggestions after looking over the code:

- I suggest MonadBH as an alternative to MonadBloodhound (I personally prefer shorter names when they can be as distinctive as longer variants). Not a deal-breaker either way, though.

- I noticed some field names that are not idiomatic in the Haskell ecosystem, like build_timestamp. The usual version would be buildTimestamp.

- I suggest making BH a newtype instad of a synonym. This way, we would have a meaningful runBH command exported that would have a more sensible/practical (flipped) arg order and wouldn't require a MonadReader import to function. It would also be a MonadTrans. I would still keep the current ReaderT BHEnv instance, as that's pretty straightforward. Signatures for BH would then be:

```haskell
newtype BH m a = ...
runBH:: BHEnv -> BH m a -> m a

example = runBH env $ do
   ...
   ...
```

- Unrelated suggestion: I recommend manually exporting the functions from the Types module while grouping them under sensible headlines. 

--- Operational Types
[MonadBloodhound, et. al would go here]
--- Server Types
--- Query Types
--- etc.

I'm not fluent enough with TS to say what those groupings should be, but I think it would help docs become more readable. Haddock page for Types is pretty hard to sift through right now.

MichaelXavier:
## MonadBH

Agreed. Made that change. I don't think the more verbose name was
doing us any favors.

## Field Names

I know we're already making breaking changes here and agree that
camelcased fields would be more idiomatic, but I will defer to
@bitemyapp for how much breakage should go into this feature. Again, I
suspect we'll be making a batch of changes like this probably in a
fork and it would be most sensible to hold this feature back until we
get some mileage on those changes so we only have to rip the band-aid
off once.

## Documentation

Agreed. I may take a stab at this. I think its useful to segment the
haddocks for the end user.

## Transformer

I've had type aliases like this blow up in my face before so I tend to
agree about the newtype. I note in your comment you parameterize the m
rather than fixing it to IO. I think this is worth us discussing:

Positives of using IO:
- Simple to understand type
- Don't have to provide lots of typeclass instances for our stack.

Negatives to using IO:
- Hard to fit calls into bloodhound into existing application
  transformer stacks except as atomic calls that you lift. This means
  you either lose the benefit of pipelining or have to constantly
  reconstruct the stack to run it (this is probably more a line noise
  issue than a perf one since its newtypes all the way down).

Positives to using m:
- Makes the functions more generic and easier to use in contexts we
  haven't thought of.
- We get to have a nicer runBH with the config first which is almost
  always easier to use.

Negatives to using m:
- For convenience we'll want to use GeneralizedNewtypeDeriving to
  derive instances for the mtl typeclasses, e.g. deriving (MonadState
  s) so if your inner m has that, you don't have to change your code
  if it happens to be in a MonadBH. This isn't hard work its just
  boilerplate.
- I imagine we'd want a MonadReader that sticks to the inner monad's
  reader if available, the goal being least conflict.

I've just started getting into writing transformers as a library
author with an eye towards interop in "Real Applications", so I don't
have a good sense of what the best choice is. Would love for you folks
to weigh in.

bitemyapp:
@MichaelXavier feeling more strongly about `IO` vs `m`? Got anything @ozataman? The stack churn using `IO` seems less than ideal.

MichaelXavier:
I'm pretty happy with the latest changes going with `m` over `IO`. We're currently using this code in production along with additional changes which eliminate the use of String in favor of Text. Things seem to be going well. As far as I can tell, these are the only changes we were going to make in the short term. Will ping Oz independently to see if he has any final comment and if not I'll submit PRs on those features.

bitemyapp:
@MichaelXavier That sounds good then. I figured y'all would test it in prod, so I'm satisfied this is good then.

Go ahead and include the String -> Text changes if it wouldn't be too much hassle. Please update the docs as well, which'll be particularly important given the MT stack now.

Thanks for everything, let me know when it's ready to merge.

MichaelXavier:
Ok. I've got that merged in. Unfortunately, it looks like the dependency situation for doctest is such that dependencies will fail to resolve on ghc 7.6, only for the doctest build target. I think the thing there is doctest forces you to go with the dependencies that ship with the compiler, so we get stuck at a version of bytestring too old for `uri-bytestring`. I've put doctests under a conditional that won't build on < ghc 7.8. This sucks but at least we get the rest of the test suite on 7.6 and everything on 7.8+. Please look this over and merge if you like it.

bitemyapp:
@MichaelXavier do you still use GHC 7.6 for anything? I'd heard of people using it for the faster builds, but it seems to me like GHC 7.10 is faster than 7.8 so perhaps at least that purpose can be replaced. Would it make sense to drop 7.6 support with this major release?

MichaelXavier:
We actually do still use 7.6 but we're pretty close to getting off of it. To clarify though, the project builds and works on 7.6. We've got it running in production under 7.6. I think doctest is the challenge here with 7.6, and I'm not sure yet what's going on the travis build. I wonder if we should use a sandbox for the build. Its totally building on my machine.

MichaelXavier:
Sorry for dropping the ball on this. This week I'm going to try to figure out why travis is rejecting this build. I'd like to get this in sooner rather than later as these changes will be really helpful to industrial users of bloodhound.

bitemyapp:
@MichaelXavier sounds good :) :+1:

MichaelXavier:
It turns out the issue is that I *said* i was going to disable doctests with a flag for 7.6 but actually disabled regular tests and left doctests on. Looks like the tests passed. Let me know if anything else is needed.

bitemyapp:
@MichaelXavier this looks good but what of the documentation in the README?

MichaelXavier:
Ok I've updated the docs. A couple points of order:

1. How would you feel about a newtype for From and Size on search. Technically, I would normally put these under smart constructors but that doesn't seem to be a convention right now for this project so maybe that's going a bit too far.
2. How do you like to do contributor recognition for the contributors thus far? I've seen anything from co-authorship in the cabal file, to a section in the readme, to a contributors file that ships with the sdist?

bitemyapp:
@MichaelXavier 1. I'm open to a newtype or a smart constructor to firm up the API there. Your choice.

2. Prefer cabal file and/or README section to the latter option. You can make the call on how to list yourself.

MichaelXavier:
Ok I think that's the last of my earmarking! Thanks for the patience.

bitemyapp:
@MichaelXavier does anything further need to happen for https://github.com/bitemyapp/bloodhound/tree/next-gen or can I merge to master and start thinking about a release with a major version bump?

MichaelXavier:
The only things I had on my notes I would not consider blockers.

1. Was going to ask you what you thought about a Monoid instance for SearchHits. Wasn't entirely sure yet if a maximum (max score) is a law abiding monoid. Sometimes its useful to have a mempty if you want to short-circuit a search as a noop. This is certainly not something that should hold up release though. Let me know the best channel to discuss. Maybe another issue?
2. Eventually it may be wise for us to research the error conditions for ES. Right now the end user would probably catch HttpException, but we'll eventually want to put those into a discrete type specific to ES and also see if there's cases where the body contains an error. Again, that shouldn't hold up release.

PR #29, opened 2015-03-12
Adjusts the format of the OR filter

bnordbo:
Only verified on ES 1.2.1, but the fix is similiar to caecf0fe36ee12c26067081356269f45c11fe2fc.


bitemyapp:
The test shows green across all versions, so I'm merging this. Thank you!

PR #33, opened 2015-05-13
make score nullable

MichaelXavier:
This is based off nex-gen and changes the type of Score, which can sometimes be null.


bitemyapp:
Cheers :)

:bear: :dog:

PR #34, opened 2015-05-14
Add Monoid instance for Search Hits

MichaelXavier:
Useful for combining search result sets or defaulting when a search
can't be performed


bitemyapp:
```
 Couldn't match type `Maybe a0' with `Double'
    Expected type: Score
      Actual type: Maybe a0
```

bitemyapp:
Awesome, thank you :) :+1: :cat2:

MichaelXavier:
Emoji game on point.

PR #43, opened 2015-07-09
Add Query Filter

bnordbo:
This adds the [Query Filter](https://www.elastic.co/guide/en/elasticsearch/reference/1.6/query-dsl-query-filter.html) filter. Tested ok on ElasticSearch 1.5.2, and it ought to work on all versions since 0.90.


bitemyapp:
@bnordbo so that we can track compatibility, would you please add a test?

bnordbo:
@bitemyapp, of course. I added a separate test for caching enabled since the serialization format is different. Let me know if you would rather see the tests merged into one.

bitemyapp:
@bnordbo poifekt. Thank you! :bear: :honey_pot: :smile:

PR #45, opened 2015-07-10
44 versioning

MichaelXavier:
Turns out that there is a path for rendering queries in http-types. Also turns out that I hadn't run the Enum tests and they had some erros which I fixed. Working on getting Travis to build on our fork but I think it should build this when there's a PR. Please don't hesitate to let me know if there's anything you're iffy on. I don't want our immediate need to rush something unfinished out the door.

### DIFF
--- a/README.md
+++ b/README.md
@@ -225,7 +225,7 @@ instance FromJSON Location
 -- λ> encode $ Location 10.0 10.0
 -- "{\"lat\":10,\"lon\":10}"
 
-resp <- withBH' $ indexDocument testIndex testMapping exampleTweet (DocId "1")
+resp <- withBH' $ indexDocument testIndex testMapping defaultIndexDocumentSettings exampleTweet (DocId "1")
 
 ```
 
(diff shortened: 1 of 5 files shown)


MichaelXavier:
I think it was but it was too lazy. evaluate forced the error to be evaluated.

bitemyapp:
@MichaelXavier ahhh, I'll have to keep the strictness difference in mind. So they're the same…modulo bottom :laughing: 

This looks good modulo docs, do the examples need updated?

MichaelXavier:
All the doctest ones are fixed. Just fixed the last instance I could find in the README.

MichaelXavier:
ooh, one last thing, EsResult should return a Version. Let me go ahead and implement that.

bitemyapp:
@MichaelXavier thank you so much! This is great :)

I can't guarantee this'll go up immediately – v. slow internet at home, but I'll try to find an opportunity this weekend sometime.

PR #51, opened 2015-07-17
Fixed mapping API

### LINKED ISSUE #49: Incorrect mapping API is used
According to the [docs](https://www.elastic.co/guide/en/elasticsearch/reference/1.6/indices-put-mapping.html), the proper URL to call when installing a mapping is `/INDEX/_mapping/MAPPING_NAME`. This is true for protocol versions 1.6, 1.5, 1.4, 1.3 and 1.2, which are tested for in `.travis.yml`. However, the URL being called by putMapping (et al.) is `/INDEX/MAPPING_NAME/_mapping`, which was last seen in protocol version 0.90.

This occurs for [putMapping](https://github.com/bitemyapp/bloodhound/blob/1db8ad7ca08ec619fd36a1945b0a1d714055def7/src/Database/Bloodhound/Client.hs#L315) and [deleteMapping](https://github.com/bitemyapp/bloodhound/blob/1db8ad7ca08ec619fd36a1945b0a1d714055def7/src/Database/Bloodhound/Client.hs#L329).

centromere:
Fixes #49.

### DIFF
--- a/src/Database/Bloodhound/Client.hs
+++ b/src/Database/Bloodhound/Client.hs
@@ -312,7 +312,7 @@ putMapping :: (MonadBH m, ToJSON a) => IndexName
                  -> MappingName -> a -> m Reply
 putMapping (IndexName indexName) (MappingName mappingName) mapping =
   bindM2 put url (return body)
-  where url = joinPath [indexName, mappingName, "_mapping"]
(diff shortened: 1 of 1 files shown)


bitemyapp:
Thank you *again* :bear: :honey_pot: :grinning:

PR #53, opened 2015-07-20
Scan type

bermanjosh:


bloodhound.cabal
```diff
@@ -1,5 +1,5 @@
 name:                bloodhound
-version:             0.6.0.1
+version:             0.6.0.2
```
bitemyapp: Sorry to bikeshed, but probably `0.6.1.0` since it's not merely a bug fix.

I usually use this as a reference when I need to remind myself how to bump versions (which is often – I am very forgetful): http://semver.org/

src/Database/Bloodhound/Types.hs
```diff
@@ -604,7 +606,16 @@ data Search = Search { queryBody       :: Maybe Query
...
+                     , size            :: Size 
+                     , searchType      :: SearchType } deriving (Eq, Show)
+
+data SearchType = SearchTypeQueryThenFetch 
+                | SearchTypeDfsQueryThenFetch
+                | SearchTypeCount
+                | SearchTypeScan
+                | SearchTypeQueryAndFetch
+                | SearchTypeDfsQueryAndFetch
+  deriving (Eq, Show)
```
bitemyapp: This is reasonable and I won't block on this, but I get a bit suspicious when a product-of-sum and maybe fields are combined. Usually means you want to lift the sum up a level and case out when fields will or will not occur in the variants.

bitemyapp:
@bermanjosh awesome, thank you! :+1: :fireworks:

bermanjosh:
sorry it dragged out so much

PR #54, opened 2015-07-20
Fixed IndexTemplate serialization bug

centromere:
Previously, when IndexTemplates were serialized they would look like this:

    {"settings":{"settings":{...

due to an oversight on my part. This PR fixes that.


bitemyapp:
@centromere danke. :+1:

Issue #57, opened 2015-07-23
Eliminate partiality in ESResult


MichaelXavier:
The `ESResult` type is described as the structure for when a query is successful. However, in the case where `found` would be False (i.e. an ID lookup that doesn't exist), only the `_index`, `_type`, `_id`, and `found` fields are present. From a practical matter, to parse the found=false case without partiality, you'd have to parse it into a raw `Value`, assert its an `Object` and then lookup the found field and do something different if its false, because your JSON parse will fail if it is.

I'm not sure if this is the best solution but I propose `ESResult` to have a field `foundResult :: Maybe (ESFoundResult a)`, which would take up the fields `_version` and `_source`. You could even provide a backwards compatibility function `found :: ESResult -> Bool; found = isJust . foundResult`. This idea only holds together if there aren't further wacky permutations of ESResult that ES may send back, but I haven't encountered such cases yet.

bitemyapp:
This sounds reasonable. I don't mind doing a major version bump for this.

bitemyapp:
This was fixed when your PR landed, closing.

PR #58, opened 2015-07-27
Eliminate partiality in EsResult

MichaelXavier:
This will affect #57. Tests run on my machine but we shall see what travis says.


bitemyapp:
Travis is satisfied -> :bear: is satisfied.

:+1: :bow:

PR #59, opened 2015-07-28
Fixed bug with IndexSettings serialization

centromere:
According to the [documentation](https://www.elastic.co/guide/en/elasticsearch/reference/current/indices-update-settings.html), `IndexSettings` are serialized as `index.number_of_[shards|replicas]`, not merely `shards|replicas`.

### DIFF
--- a/src/Database/Bloodhound/Client.hs
+++ b/src/Database/Bloodhound/Client.hs
@@ -91,7 +91,7 @@ import           Database.Bloodhound.Types
 -- >>> let runBH' = withBH defaultManagerSettings testServer
 -- >>> let testIndex = IndexName "twitter"
 -- >>> let testMapping = MappingName "tweet"
--- >>> let defaultIndexSettings = IndexSettings (ShardCount 3) (ReplicaCount 2)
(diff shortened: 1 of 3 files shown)


centromere:
The CI test failed due to the test taking longer than 10 minutes to run. It took too long to run because ElasticSearch was operating slowly on account of the `defaultIndexSettings` specifying three shards and two replicas. If you specify one shard and no replicas like so:
```haskell

createExampleIndex = createIndex (IndexSettings (ShardCount 1) (ReplicaCount 0)) testIndex

```

then the test runs much faster. All previous tests did not suffer from this problem because the shards/replicas were never specified correctly to begin with. This is confirmed in the ES logs:

`[2015-07-28 19:53:09,181][INFO ][cluster.metadata         ] [Bloodsport] [twitter] creating index, cause [api], templates [], shards [5]/[1], mappings []`

bitemyapp:
@centromere Please do go ahead and change the tests to use 1 shard and 0 replicas so the CI passes.

centromere:
Done.

bitemyapp:
@centromere thank you again! :+1: :dog: :fireworks:

PR #60, opened 2015-07-28
Add value_count aggregation support

MichaelXavier:
This also adds a newtype for script. I would recommend the "Either Text
Text" for TermsAggregation be instead Either FieldName Script, but that
isn't directly related to this PR.


bitemyapp:
Good catch w/ `TermsAggregation`!

bitemyapp:
Danke :de: :beers: :+1: :smile_cat:

PR #63, opened 2015-08-18
url query encoding bug

bermanjosh:
@bitemyapp Sorry, I missed a small bug based on the different way the query parameter renderer worked

### DIFF
--- a/bloodhound.cabal
+++ b/bloodhound.cabal
@@ -1,5 +1,5 @@
 name:                bloodhound
-version:             0.7.1.1
+version:             0.7.2.0
 synopsis:            ElasticSearch client library for Haskell
 description:         ElasticSearch made awesome for Haskell hackers
 homepage:            https://github.com/bitemyapp/bloodhound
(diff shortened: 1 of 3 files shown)


bitemyapp:
@bermanjosh thank you for catching this, but could you write a test that exercises the bug you're fixing please? I don't want a regression to occur.

bermanjosh:
@bitemyapp I'm sorry, but I have no idea why CI is failing. The log says it's tripping `hspec-1.12.4`. I thought that's excluded in the cabal file.

bitemyapp:
@bermanjosh probably TravisCI is fucking us. Now we do the dance of death where I swat the .travis.yml and bounce your build until it works. Sorry mate, I'll ping you if I run into a non-spurious error. :|

bitemyapp:
@bermanjosh I'm just going to merge it and see what happens.

bermanjosh:
@bitemyapp Good to know I'm not going crazy - let me know if I missed something (again...).

bitemyapp:
@bermanjosh https://travis-ci.org/bitemyapp/bloodhound/builds/76873638

Nope, you're in good shape. The builds use Stack instead of cabal-install now.

PR #64, opened 2015-08-26
export BH constructor

MichaelXavier:
Turns out if you have to implement instances (orphan or otherwise) for BH, it is pretty useful to export the full BH constructor. I can't really see any harm in exporting it other than if we at some point made a breaking change to the underlying value and were going to make it API-compatible by converting in runBH, but people would be using runBH anyways and would only be using the constructor for instance declarations/derivations.


bitemyapp:
Dunzo, thank you :) :+1:

Issue #65, opened 2015-08-31
MonadMask / bracket


bitemyapp:
Should BH have a MonadMask instance? What's the right thing to do here?

/cc @MichaelXavier

MichaelXavier:
You may consider if it incurs additional dependencies. It may already indirectly require the package and if so, its not too big a deal IMO. You would probably just newtype derive an instance, right?

mwotton:
this came up in the context of wanting to do bracket stuff inside the monad transformer stack. i ended up just using Control.Exception, but it's not a tremendously satisfying solution, as it means you have to do everything at the IO level.

MichaelXavier:
PR is up for your perusal.

bitemyapp:
Closed by #67

PR #67, opened 2015-09-05
Take a stab at deriving Monad(Throw|Catch|Mask)

MichaelXavier:
Took a stab at #65 and it didn't require me to add any direct dependencies.


bitemyapp:
/cc @mwotton - Does this catch all the instances you wanted? I believe it does :)

Thank you @MichaelXavier :dart:

mwotton:
yep, looks good. sorry for the slow response.

bitemyapp:
@mwotton @MichaelXavier thank you both :) :bear:

Issue #62, opened 2015-07-29
Impossible to get child documents


centromere:
As it stands, it is impossible for a user to retrieve/test for the existence of individual child documents, since the parent ID must be specified (`?parent=...`). This is needed so that ES can route the request to the proper shard. How should the API be altered to accommodate this?

centromere:
For my purposes I've changed the type signature of `documentExists` to be as follows:
```haskell
documentExists :: MonadBH m => IndexName -> MappingName
               -> Maybe DocumentParent -> DocId -> m Bool
```
It works, but there are other ES features which, if support is desired, would necessitate an API change. Perhaps something like a `GetDocumentSettings` type should be used?

PR #72, opened 2015-09-15
Added parent support to documentExists

centromere:
This may not be the best way to approach the problem. See #62.


bitemyapp:
@centromere could you add a test that exercises this?

centromere:
I've added a test.

bitemyapp:
@centromere thank you :)

PR #73, opened 2015-10-07
Fix for #38

### LINKED ISSUE #38: Provide safety by using URL-encoding
Hi

Thank you for publishing this nice library.

Wrapped in its own type, I expected `DocId` to safely handle any input. However, `indexDocument` throws an `InvalidUrlException` at me because I've got slashes and spaces in there. Would you please add automatic URL encoding?

If doing this by default really hampers with performance, could we at least expose encoding functionality as courtesy to users?

MaxDaten:
It's a just a proof of concept to fix #38. I can't guarantee that it will behaves in all valid case exactly as before, although all test-cases are green ;) Feel free to adopt the changes or to dump it.


bitemyapp:
Happy to roll with it, thank you @MaxDaten!

:+1: :bear: :honey_pot: :smiley:

MaxDaten:
Your welcome!
:palm_tree:

PR #74, opened 2015-10-08
Add readme and changelog to extra source files

MichaelXavier:
Generally seems like good practice for the readme but there's a clear
benefit to adding changelog.md to extra-source-files: it gets parsed by
hackage and linked right from the haddocks!


bitemyapp:
Ah :nut_and_bolt: :shipit: :nut_and_bolt:, good catch. Thank you!

PR #75, opened 2015-10-09
Add a type for errors (for your consideration)

MichaelXavier:
Don't merge this yet. Its a discussion point

We still should discuss having a either a sum type or other affordance
for the case of parsing a response body which could be an error. You
can't just simply parse Either ESError a because aeson for some reason
provides an instance for (FromJSON a, FromJSON b) => FromJSON (Either a
b) which is highly Haskell-centric.

Bloodhound currently defines some convenient wrappers for happy path responses but nothing for error cases. As best I can tell, they have a single format for errors: a code (which seems to be redundant with the HTTP status code) and a big ugly error message. I don't think we should go down the path of trying to parse the message, but the type I provide here is enough to do some simple pattern matching on status which is about all you can hope for. The remaining question would be how much support would we want to add to bloodhound towards making this type easy to use. I can see a few approaches:

1. None. Take the patch as is. User is invited to use Alternative,  MaybeT, or something else when parsing the response body to try to parse to an error. This is the most flexible but least user friendly. The user will have to dig around the docs to find an error type and then figure out how to use it.
2. Add a newtype over `Either` that can parse the desired type or an error, e.g. `newtype EsErrorable a = EsErrorable (Either EsError a)`. The user is then free to parse responses like EsErrorable (EsResult a) if they want to catch errors at the type level. This is decently flexible and gives you a batteries-included way to parse errors. If we go with this one we've got to come up with a better name than I have above.

There may be other ways as well. LMK what you think.

### DIFF
--- a/src/Database/Bloodhound/Types.hs
+++ b/src/Database/Bloodhound/Types.hs
@@ -61,6 +61,7 @@ module Database.Bloodhound.Types
        , Reply
        , EsResult(..)
        , EsResultFound(..)
+       , EsError(..)
        , DocVersion
        , ExternalDocVersion(..)
        , VersionControl(..)
@@ -416,13 +417,18 @@ data EsResult a = EsResult { _index      :: Text
                            , _id         :: Text
                            , foundResult :: Maybe (EsResultFound a)} deriving (Eq, Show)
 
-
 {-| 'EsResultFound' contains the document and its metadata inside of an
     'EsResult' when the document was successfully found.
 -}
 data EsResultFound a = EsResultFound {  _version :: DocVersion
                                      , _source   :: a } deriving (Eq, Show)
 
+{-| 'EsError' is the generic type that will be returned when there was a
+    problem. If you can't parse the expected response, its a good idea to
+    try parsing this.
+-}
+data EsError = EsError { errorStatus  :: Int
+                       , errorMessage :: Text } deriving (Eq, Show)
 
 
 {-| 'DocVersion' is an integer version number for a document between 1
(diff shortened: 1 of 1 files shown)


bitemyapp:
>As best I can tell, they have a single format for errors: a code (which seems to be redundant with the HTTP status code) and a big ugly error message.

That's all I've seen as well.

>I don't think we should go down the path of trying to parse the message

I don't either, that's a "tracking upstream Hell" quagmire waiting to happen.

Patch as-is seems okay and like a good improvement, I want to try to understand option 2.

>Add a newtype over Either that can parse the desired type or an error, e.g. newtype EsErrorable a = EsErrorable (Either EsError a).

Is this somehow different from the aforementioned quagmire? Maintaining a parser that understands their error message formats and maintaining that across the (how many ES versions is it now?) various ES releases doesn't sound like fun.

>The user is then free to parse responses like EsErrorable (EsResult a) if they want to catch errors at the type level. This is decently flexible and gives you a batteries-included way to parse errors.

Apologies for my hyper-density, is this part of the, "we're going to try to parse errors" approach?

Cheers and thanks for kicking the can with me @MichaelXavier :)

MichaelXavier:
To clarify, `newtype EsErrorable a = EsErrorable (Either EsError a)` would use the current, vague `EsError` presented in this patch, the one that carries an `Int` status and an opaque `Text` error. The only thing it would provide is a parseJSON that would be something like `parseJSON v = (Right <$> parseJSON v) <|> (Left <$> parseJSON v)` That would consequently be the only work that we'd be saving users from. So I guess the real question is is that convenience worth the newtype or do we just keep `EsError` and leave the alternating parsers as an exercise to the user or as a suggestion in the docs.

bitemyapp:
@MichaelXavier do you have a preference or inclination on this?

MichaelXavier:
Having recently written a *function* to parse either a or EsError, it wasn't difficult. I think the more conservative choice is to just have the EsError type. It would be easy to add a fuller implementation if there was a more compelling reason later.

bitemyapp:
@MichaelXavier comports with my preferences. Thank you sir :crown:

bitemyapp:
@MichaelXavier seems like this would bear documenting and a test. Up for it?

MichaelXavier:
@bitemyapp certainly. Will send another PR.

bitemyapp:
@MichaelXavier thank you :)

PR #77, opened 2015-10-17
Add test for error parsing

MichaelXavier:
So here's the test. The type has a docstring on it. Anything else you had in mind or should this do it?


MichaelXavier:
I'm not sure what's going on with the test suite but my test is green...

bitemyapp:
@MichaelXavier you're fine fine, doctests are squawking and the 7.8 yaml needs updated for the new lower bound.

PR #78, opened 2015-10-19
Update index settings

MichaelXavier:
Let me know what you think. For the most part this is a rote translation of ES' docs, I have no idea what a lot of these settings actually do. In some cases the docs gave no guidance as to what the JSON should actually look like so I've gathered the types from looking at stack overflow. I've improved type specificity where sensible (NominalDiffTime for time intervals, NonEmpty where applicable).

Let me know what you think and if you think anything else is needed. I didn't really have any ideas for exhaustive testing of that since it has to go to a live server to take effect.

### DIFF
--- a/src/Database/Bloodhound/Types/Class.hs
+++ b/src/Database/Bloodhound/Types/Class.hs
@@ -3,9 +3,9 @@ module Database.Bloodhound.Types.Class
        ( Seminearring(..) )
        where
 
-#if !MIN_VERSION_base(4,8,0)
+-- #if !MIN_VERSION_base(4,8,0)
 import Data.Monoid
-#endif
+-- #endif
 
 class Monoid a => Seminearring a where
   -- 0, +, *
--- a/src/Database/Bloodhound/Types/Class.hs
+++ b/src/Database/Bloodhound/Types/Class.hs
@@ -3,9 +3,9 @@ module Database.Bloodhound.Types.Class
        ( Seminearring(..) )
        where
 
--- #if !MIN_VERSION_base(4,8,0)
+#if !MIN_VERSION_base(4,8,0)
 import Data.Monoid
--- #endif
+#endif
 
 class Monoid a => Seminearring a where
   -- 0, +, *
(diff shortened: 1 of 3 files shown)


src/Database/Bloodhound/Types/Class.hs
```diff
@@ -3,9 +3,9 @@ module Database.Bloodhound.Types.Class
        ( Seminearring(..) )
        where
 
-#if !MIN_VERSION_base(4,8,0)
+-- #if !MIN_VERSION_base(4,8,0)
```
bitemyapp: unnecessary? unhelpful?

bitemyapp:
I find the under-documentation and less than stellar testability troubling as well, but we've been playing the [chinese room game](https://en.wikipedia.org/wiki/Chinese_room) with Elasticsearch for awhile now.

Way to validate this stuff would be to treat the server like a state machine, run tests that update index settings, then pull the current state down from the metadata APIs and match against expectations.

screenshot from 2015-10-19 11 59 19

Finally fixed the 7.8 build :cry: 

I've bounced the PR/push builds on Travis. In the meantime, how do you feel about adding tests along the lines I mentioned? Do you think it's worth it? I don't want stuff like this to silently break --- could affect people in production.

bitemyapp:
@MichaelXavier looks like a legit type error (from your PR, not from my stuff)

src/Database/Bloodhound/Types/Class.hs
MichaelXavier: Whoops. My bad. I had done that on a different day to get haskell interactive process and stack to play nice and forgot to put it back.

MichaelXavier:
I think you're probably right on testing. If I can get the server to echo back the settings to me, then maybe the plan should be to contrive 1 or more settings updates that run the gamut, then get the server to repeat it back to us. The fact that its confined to an index and I can create one just for this makes me think this could work.

bitemyapp:
@MichaelXavier coolio, sounds copacetic then. Thank you sir :)

Issue #79, opened 2015-10-23
Add support for index aliases


MichaelXavier:
I'm knee deep in this one right now so this is a placeholder. This will add support for creating/updating index aliases as well as reading them back.

One thing important to note is this idea of adding these PUT/POST APIs and then adding the corresponding GET for testing and general use is that bloodhound may no longer get to punt on parsing the response. Most ops now are updates so they just return `m Reply`. For an API like say `getIndexAliases` whose sole purpose is to return a well-typed response, it seems unsatisfying to return an `m Reply` and hint in the docs that you should try parsing it as the right type.

The temporary approach I'm taking while working on this is to return `m (Either EsError TheThingYouExpect)`. It will try to parse the thing first and failing that, will parse an EsError. AFAIK all ElasticSearch errors conform to the EsError format, so misses on this should be rare. If it can parse neither, it throws a `StatusCodeException` from network client.

So perhaps mull over this approach and see what you think about it. It will also be relevant to that index settings ticket in #76. We'll probably use some subset of the index aliases feature in a Real App before I submit the PR.

MichaelXavier:
The feature is basically done at this point but I have some cleanup and possibly some more thorough integration testing to do. I actually have Query/Filter and all the types they use with ToJSON/FromJSON instances now that are tested against eachother with quickcheck. This actually caused me to discover a few minor bugs with the JSON bloodhound is producing today. I've had to make some minor changes to types to squeeze out some ambiguous cases. Once I get this thing more polished we can discuss if these breaking type changes are worth it. They are all very simple to update and gravitate towards more precise types.

Now the weird part: I'm using stack and just recently the doctests will build but not run. It looks like its using 2 different versions of http-types in the doctests runtime and I don't know what even the fuck. Was hoping you may know what's going on @bitemyapp 

```
stack test   
bloodhound-0.9.0.0: unregistering (dependencies changed)
bloodhound-0.9.0.0: configure
Configuring bloodhound-0.9.0.0...
bloodhound-0.9.0.0: build
Preprocessing library bloodhound-0.9.0.0...
[1 of 4] Compiling Database.Bloodhound.Types.Class ( src/Database/Bloodhound/Types/Class.hs, .stack-work/dist/x86_64-linux/Cabal-1.22.4.0/build/Database/Bloodhound/Types/Class.o )
[...]

PR #80, opened 2015-11-03
Index aliases

MichaelXavier:
This is in reference to #79 and is meant for review. Apologies for the size and duration of the PR but most of the work was due to me pulling the string on the fact that aliases can include filters, which can include queries, so I had to add FromJSON instances for all of that.

A couple discussion points:
1. `getIndexAliases` seems like its the first case where its purely a GET operation with a non-trivial response type. I went with a response type of `Either EsError IndexAliases` and exported the utility function `parseEsResponse` to capture the case that all JSON responses should either parse as the intended type or EsError. Do note that this function does throw in the (AFAIK impossible) case that the response doesn't parse as either. At least it documents what it throws...
2. There are some breaking changes to the types. I chose to test the JSON instances via quickcheck and caught some bugs and ambiguities that the types allowed. A lot of them are from combining lists and maybe, where there is no semantic difference between `Nothing` and `Just []`. In those cases I updated it to `Maybe (NonEmpty a)`. The full list of broken changes can be provided in the changelog if desired. These breaking changes should be really easy to fix for end users though, especially since Bloodhound uses NonEmpty in other places.
3. I get some doctest errors on my machine. We'll see if travis has the same issue.

### DIFF
--- a/bloodhound.cabal
+++ b/bloodhound.cabal
@@ -45,7 +45,8 @@ library
                        data-default-class,
                        blaze-builder,
                        unordered-containers,
-                       mtl-compat
+                       mtl-compat,
+                       hashable
   default-language:    Haskell2010
 
 test-suite tests
@@ -68,7 +69,9 @@ test-suite tests
                        vector,
                        unordered-containers >= 0.2.5.0 && <0.3,
                        mtl,
-                       quickcheck-properties
+                       quickcheck-properties,
+                       derive,
+                       quickcheck-instances
   default-language:    Haskell2010
 
 test-suite doctests
--- a/src/Database/Bloodhound/Client.hs
+++ b/src/Database/Bloodhound/Client.hs
@@ -287,7 +287,7 @@ parseEsResponse :: (MonadBH m, MonadThrow m, FromJSON a) => Reply
 parseEsResponse reply
   | respIsTwoHunna reply = case eitherDecode body of
                              Right a -> return (Right a)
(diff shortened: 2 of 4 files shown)


bitemyapp:
@MichaelXavier I tink you made it mad.
screenshot from 2015-11-02 22 01 14

MichaelXavier:
Woah. That's bizarre. I have no idea how my tests built, ran, and passed without typeable...

MichaelXavier:
Ayyyy. Tests pass!

bitemyapp:
>Woah. That's bizarre. I have no idea how my tests built, ran, and passed without typeable...

@MichaelXavier 7.8 vs. 7.10? I don't know.

Are you comfortable with this? If so, I'll merge and try to cut a release before I fly out for Hac-Phi.

MichaelXavier:
I'm comfortable with it. Don't worry too much about cutting a release. I don't know if anyone is clamoring for this feature. I'll likely next circle back to my other PR and fix it up as well. For our own purposes, we're using our own fork so we're not affected by a later release.

bitemyapp:
@MichaelXavier nah just opportunism. I try to have at least one new thing for the releases I do for Stackage if possible.

MichaelXavier:
Ok, just keep in mind if you are going to cut a release that this merged to the index-aliases branch of this repo so you'll need to merge that to master when ready.

bitemyapp:
@MichaelXavier thank you for the warning, I am a bit slow.

Issue #76, opened 2015-10-13
Update Index Settings


MichaelXavier:
I've hit a case where it would be useful to have this API so this is an issue do discuss the possible implementation. The documentation is refreshingly detailed about updating index settings:

https://www.elastic.co/guide/en/elasticsearch/reference/current/indices-update-settings.html

My first thoughts on a plan:
1. We can't use `IndexSettings`. The docs show a much more rich set of options that can be tweaked than `IndexSettings` provide.
2. You will get an error from the server if you provide an empty set of updates. Somewhere in the update type it should have a `NonEmpty IndexSettingsUpdate`.
3. `IndexSettingUpdate` would be a sum type over all the settings shown.
4. The type would then be something like `updateIndexSettings :: MonadBH m => NonEmpty IndexSettingUpdate -> IndexName -> m Reply`

Let me know what you think of the plan.

bitemyapp:
@MichaelXavier zero things to object to in the proposal you've laid out here. Go for it!

One thing that caught me eye when I was reading the docs, the flipping around of index settings (refresh rate) for bulk indexing. Wondering how well that works and how dangerous it is. I need some AWS budget to play with :P

PR #82, opened 2015-11-14
Date range agg

MichaelXavier:
Excuse the giant diff here, I may be bad at PRs. I based my PR off of the as-yet-unmerged low-level-scroll PR, so the diff should shrink significantly after that merge.

This adds the date range aggregation type feature.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,11 +1,11 @@
-# NB: don't set `language: haskell` here
+sudo: false
 
-# See also https://github.com/hvr/multi-ghc-travis for more information
+addons:
+  apt:
+    packages:
+      - libgmp-dev
+      - openjdk-6-jre
 
-# The following lines enable several GHC versions and/or HP versions
-# to be tested; often it's enough to test only against the last
-# release of a major GHC version. Setting HPVER implictly sets
-# GHCVER. Omit lines with versions you don't need/want testing for.
 env:
  # - GHCVER=7.6.3  ESVER=1.6.0 # Deprecated
  # - GHCVER=7.8.3  ESVER=1.0.3 # Deprecated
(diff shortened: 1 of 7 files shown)


bitemyapp:
@MichaelXavier does this pull in index aliases stuff as well?

MichaelXavier:
Ah looks like it does. I'll go ahead and get that PR fixed. Then the merge order should be index aliases -> low level scroll -> date range agg.

bitemyapp:
@MichaelXavier I'd be fine merging one big mother PR into master. Saves me bookkeeping :P

MichaelXavier:
Fair enough. Since its already in here I'll just use this branch as a start point. AFAICT the only things I need to do are loosen up the "errors" package in test to get tests running again and then add some integration tests for index settings update. @bitemyapp any reason not to just merge those to master? I think the trip up with the huge diff was that I forgot that my index-aliases branch didn't go to master but this PR is targeting master.

bitemyapp:
@MichaelXavier merged index-aliases followed by low-level scroll, I think. Let me know if you need me to do anything else! Thank you as always :bear:

bitemyapp:
@MichaelXavier https://travis-ci.org/bitemyapp/bloodhound/jobs/91447777#L684 ??? Weird.

MichaelXavier:
@bitemyapp this happens when it can't parse the response even though it was successful. I suspect there was a change to the response format in that version of ES. Not sure how I'm going to check that yet...

MichaelXavier:
We've got a green test suite! Merge at will.

There was 1 semantic change but I think its for the better. The issue was that specifically 1.4.1 returned a different alias format than any other version, including 1.4.5. This was easily accounted for. I couldn't figure this out from the "impossible" exception case of bloodhound raising a status code exception on a 200. You'd have to be pretty clued into the library design to know that this means it couldn't parse the body as the expected type nor as an error. I figure that this case represents an error at the protocol level and there's nothing that bloodhound can do but throw. The alternative of a cascading Either (multiple eithers) was not attractive. When something like this comes up, its either a bug in bloodhound or its a version that is not supported. Now, this hopefully rare codepath throws a custom `EsProtocolException` that includes the response body that it barfed on. In the surrounding docs the user is encouraged to submit this as a bug on the issue tracker.

If all of this sounds good, go ahead and merge and we can close #78 and #76.

bitemyapp:
@MichaelXavier Damn!  :alarm_clock: alarm_clock: alarm_clock: :fire:

Thank you as always! Between this and the Stackage pings, I'll be looking to push out a release soon.

PR #85, opened 2015-12-22
Add updateDocument

dzhus:


MichaelXavier:
This could use some tests :)

dzhus:
I added tests.

MichaelXavier:
@bitemyapp This looks good to me. One thing I'll note is that updates seem like they can also take a script with some additional parameters and configuration. I don't know anyone who uses this but if they did, we'd probably want to add an additional function to handle those updates.

bitemyapp:
@MichaelXavier the scripting stuff is gnarly and I'm mostly hoping Haskellers generally are as keen as I am to avoid it.

bitemyapp:
Thank you @dzhus ! :bear:

dzhus:
Thank you for merging this

PR #87, opened 2016-01-12
Add Read instances

dzhus:
I wanted to read a `Filter` and it kinda escalated…

### DIFF
--- a/src/Database/Bloodhound/Types.hs
+++ b/src/Database/Bloodhound/Types.hs
@@ -363,7 +363,7 @@ data Version = Version { number          :: Text
                        , build_hash      :: Text
                        , build_timestamp :: UTCTime
                        , build_snapshot  :: Bool
-                       , lucene_version  :: Text } deriving (Eq, Show, Generic, Typeable)
+                       , lucene_version  :: Text } deriving (Eq, Read, Show, Generic, Typeable)
 
 {-| 'Status' is a data type for describing the JSON body returned by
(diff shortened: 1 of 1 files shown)


bitemyapp:
I really loathe read for being partial, but I'll give in on this if @MichaelXavier gives it :+1:

MichaelXavier:
@bitemyapp there's definitely ways to use read in a way that produces a Maybe. Personally I only use it for code that isn't performance sensitive at all like CLI flags since show/read are slow and space inefficient, but I wouldn't be prescriptive about that.

I do see a bug here though that should probably be fixed: BH uses a custom `Show` instance for `TimeInterval` which as far as I can tell is only used to implement `ToJSON Interval`. There's an implication that `Show` and `Read` should round-trip, so `Show TimeInterval` should be instead derived and the `show interval` in `ToJSON Interval` should just be inlined in a where clause or something. tl;dr: derived `Read` should only be used in the presence of derived `Show`.

dzhus:
I'll take a look thanks for the comment

bitemyapp:
@MichaelXavier >there's definitely ways to use read in a way that produces a Maybe.

I know, but people don't. Sounds like you're alright with it as long as the bug is fixed, so so am I.

dzhus:
Hey guys, I've fixed the instance for `TimeInterval`, please take a look again

bitemyapp:
@dzhus works for me, thank you :)

Issue #88, opened 2016-01-20
TermsResult cannot parse from numeric or boolean JSON


MHova:
Specifically because `termKey` is declared as type `Text`, non-String bucket terms like numbers and booleans will fail parsing.

We could either keep `termKey` as type `Text` and put some more smarts in `parseJSON` or introduce more data constructors (`TextTermsResult`, `NumericTermsResult`, etc) and also put some more smarts in `parseJSON`.

bitemyapp:
@MHova I'd just as soon fix the type to be explicit about the possibility of Text, Bool, Number.

/cc @MichaelXavier

MHova:
Thanks for the quick response. Let me take a stab at fixing this.

PR #89, opened 2016-01-20
Support Bool and Numeric keys in TermsResults

MHova:
Addresses https://github.com/bitemyapp/bloodhound/issues/88

This is a backwards-incompatible change because the type of the `termKey` record in `TermsResult` and the signature of `BucketAggregation.key` have changed.

Some things I'm not sure about:
1) What is the `BucketAggregation` class used for? Is it okay to just convert `Scientific` and `Bool` keys to `Text`?
2) I'm a little concerned about the `Scientific` `Show` instance.  A JSON integer, `123` for example, will display as "123.0" as a `Scientific`. Is this acceptable?

Edit: reflect new content

### DIFF
--- a/src/Database/Bloodhound/Types.hs
+++ b/src/Database/Bloodhound/Types.hs
@@ -8,6 +8,7 @@
 {-# LANGUAGE OverloadedStrings          #-}
 {-# LANGUAGE RecordWildCards            #-}
 {-# LANGUAGE UndecidableInstances       #-}
+{-# LANGUAGE NamedFieldPuns             #-}
 
 -------------------------------------------------------------------------------
 -- |
@@ -284,6 +285,7 @@ import           Data.List                          (foldl', nub)
 import           Data.List.NonEmpty                 (NonEmpty (..), toList)
 import qualified Data.Map.Strict                    as M
 import           Data.Maybe
+import           Data.Scientific                    (Scientific)
 import           Data.Text                          (Text)
 import qualified Data.Text                          as T
 import           Data.Time.Calendar
@@ -1744,9 +1746,20 @@ class BucketAggregation a where
 
 data Bucket a = Bucket { buckets :: [a]} deriving (Show)
 
-data TermsResult = TermsResult { termKey       :: Text
-                               , termsDocCount :: Int
-                               , termsAggs     :: Maybe AggregationResults } deriving (Show)
+data TermsResult
+  = TextTermsResult
+    { textTermKey   :: Text
+    , termsDocCount :: Int
+    , termsAggs     :: Maybe AggregationResults }
+  | ScientificTermsResult
+    { scientificTermKey       :: Scientific
+    , termsDocCount           :: Int
+    , termsAggs               :: Maybe AggregationResults }
+  | BoolTermsResult
+    { boolTermKey   :: Bool
+    , termsDocCount :: Int
+    , termsAggs     :: Maybe AggregationResults }
+  deriving (Show)
 
 data DateHistogramResult = DateHistogramResult { dateKey           :: Int
                                                , dateKeyStr        :: Maybe Text
@@ -1770,7 +1783,9 @@ toDateHistogram t a = M.lookup t a >>= deserialize
   where deserialize = parseMaybe parseJSON
 
 instance BucketAggregation TermsResult where
-  key = termKey
+  key TextTermsResult{textTermKey} = textTermKey
+  key ScientificTermsResult{scientificTermKey} = T.pack $ show scientificTermKey
+  key BoolTermsResult{boolTermKey} = T.pack $ show boolTermKey
   docCount = termsDocCount
   aggs = termsAggs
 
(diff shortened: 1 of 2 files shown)


src/Database/Bloodhound/Types.hs
```diff
@@ -1744,9 +1746,20 @@ class BucketAggregation a where
 
 data Bucket a = Bucket { buckets :: [a]} deriving (Show)
 
-data TermsResult = TermsResult { termKey       :: Text
-                               , termsDocCount :: Int
-                               , termsAggs     :: Maybe AggregationResults } deriving (Show)
+data TermsResult
+  = TextTermsResult
```
bitemyapp: Please don't do this, it creates partial functions. Split the sums and records.

bitemyapp:
>1) What is the BucketAggregation class used for?

https://www.elastic.co/guide/en/elasticsearch/reference/2.x/search-aggregations-bucket.html

>Is it okay to just convert Scientific and Bool keys to Text?

I'd rather preserve the original information.

>2) I'm a little concerned about the Scientific Show instance. A JSON integer, 123 for example, will display as "123.0" as a Scientific. Is this acceptable?

There's no such thing as a JSON integer. "123" and "123.0" are the same thing as far as JSON is concerned. Elasticsearch's attempts to ignore what is expressible in JSON doesn't change this.

src/Database/Bloodhound/Types.hs
MHova: Ah yes, makes sense. Addressed in https://github.com/MHova/bloodhound/commit/b9924936228a8826806b53eb3927f249f014ff00

MHova:
1) Addressed in https://github.com/MHova/bloodhound/commit/225ef80b778e08b26626fc3da1288a1f9715192b

2) Right. The only thing that concerns me is when the server using Bloodhound responds to some browser client with JSON containing a number field of value `123.0` (when we really wanted `123`), we would now have to have some special handling somewhere to display it as `123`.

src/Database/Bloodhound/Types.hs
MHova: Ah. Thank god for tests. Rebased to https://github.com/MHova/bloodhound/commit/cba1a574c3693fa7845212393025ca48fcd4890f

bitemyapp:
@MHova  >concerns me is when the server using Bloodhound responds to some browser client with JSON containing a number field of value 123.0 (when we really wanted 123), we would now have to have some special handling somewhere to display it as 123

Well, I've tried raising this with bos, but he convinced me it's not worth exceeding the standard. ES devs were unwise to count on this, but I think you can figure your ES instance to squash/ignore the `.0`. Not ideal, but it we're [stuck in the middle](https://www.youtube.com/watch?v=DohRa9lsx0Q) anyway.

Let me know when this is ready to go forward for another round of review.

MHova:
Ha ha okay. Well my concern was no so much about ES's behavior as ES does [coercion] (https://www.elastic.co/guide/en/elasticsearch/reference/current/coerce.html) anyway. I'm more concerned about the actual displaying of this data on a front end. If my service produces a `123.0`, but we'd really like to display `123` to the user, then someone somewhere has to know to make this change.

Well in any case, yes I'm ready for another round of review.

MichaelXavier:
Since it looks like we're trying to define term keys in terms of all the non-recursive Value constructors, is it possible to have a term key that is a `Null`?

MHova:
Hmm is it even possible for a term key to come back from ElasticSearch as `null`?

MichaelXavier:
@MHova I'm not sure but can't you group term keys by data in the document. I'm curious what happens if that field in the document is `null`.

MHova:
By "group term keys" I think you're talking about the [`terms aggregation`] (https://www.elastic.co/guide/en/elasticsearch/reference/2.x/search-aggregations-bucket-terms-aggregation.html) (which, incidentally, is the very thing I'm using ElasticSearch for). `null` fields do not appear in the `terms aggregation` results. Here's a little experiment on ElasticSearch v2.1:

1)
```
PUT myindex/mytype/3
{
    "field1" : "whee",
    "field2" : null
}
```

2)
```
POST myindex/mytype/_search
{
    "size" : 0,
    "aggs" : {
        "field2terms" : {
            "terms" : {"field" : "field2"}
        },
        "field1terms" : {
            "terms" : {"field" : "field1"}
        }
    }
}
```

Result:
```
{
   "took": 4,
   "timed_out": false,
   "_shards": {
      "total": 5,
      "successful": 5,
      "failed": 0
   },
   "hits": {
      "total": 1,
      "max_score": 0,
      "hits": []
   },
   "aggregations": {
      "field1terms": {
         "doc_count_error_upper_bound": 0,
         "sum_other_doc_count": 0,
         "buckets": [
            {
               "key": "whee",
               "doc_count": 1
            }
         ]
      },
      "field2terms": {
         "doc_count_error_upper_bound": 0,
         "sum_other_doc_count": 0,
         "buckets": []
      }
   }
}
```

MichaelXavier:
@MHova gotcha. Thanks for the clarification. That was the only pot shot I could take at this. It looks good to me and I like that we're able to constrain the type to exactly what is possible.

bitemyapp:
@MHova @MichaelXavier is this good to be merged then?

MHova:
Yes sir!

bitemyapp:
@MHova :tada: :fireworks: :smile: :bear: thank you!

MHova:
Hooray!

PR #91, opened 2016-01-22
make respIsTwoHunna more semantic

MichaelXavier:
Don't mind the branch name, I got lazy.

This is actually the cause of a bug in real code. If you happen to be
using parseEsResponse (which uses respIsTwoHunna) to parse the result of
certain operations such as creating an index, those operations return a
201 and unjustly are deemed to be a failure. I will not disclose how
long this took me to figure out the true cause ;)

### DIFF
--- a/src/Database/Bloodhound/Client.hs
+++ b/src/Database/Bloodhound/Client.hs
@@ -309,11 +309,11 @@ deepMerge = LS.foldl' go mempty
         merge _ b = b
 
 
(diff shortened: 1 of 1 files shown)


bitemyapp:
>Don't mind the branch name, I got lazy.

i'm not so much of a hypocrite that I would give you guff over that mate :smile:

bitemyapp:
@MichaelXavier thank you as usual :+1:

Issue #90, opened 2016-01-21
Some errors in README


MHova:
Some lines in the README are messed up like right around this [heading] (https://github.com/bitemyapp/bloodhound#terms-aggregation). I think it's due to this commit? https://github.com/bitemyapp/bloodhound/commit/5ccffd3cc983100551a2a6a1b88ef9dc0e329e3a

MichaelXavier:
@MHova are you just talking about the errant "haskell" words left in the markdown? I just fixed those here:

https://github.com/bitemyapp/bloodhound/compare/90-markdown-bugs?expand=1

LMK if there's another issue you're talking about.

MHova:
Thanks! I noticed some other errors too:

`Term Aggregations have two factory functions, , and , and can be used as follows` (missing the names of the factory functions)

`For example, if there were a nested terms aggregation keyed to "age" in a TermsResult named , you would call` (the sentence abruptly stops here)

PR #93, opened 2016-01-24
90 markdown bugs

MichaelXavier:
This should resolve the mangling that happened when converting from org mode in #90


bitemyapp:
@MichaelXavier thank you! :tired_face:

Issue #95, opened 2016-01-28
Documentation on advanceScroll is wrong


MHova:
According to the latest ElasticSearch [docs] (https://www.elastic.co/guide/en/elasticsearch/reference/current/search-request-scroll.html), the terminating condition for scrolling is an empty `hits` array, not the lack of a `_scroll_id`

```
Each call to the scroll API returns the next batch of results until there are no more results left to return, ie the hits array is empty.
```

In fact, the previous `_scroll_id` is included in the response so if one relies on the `scrollId` field being `Nothing` as the terminating condition, one will get into an infinite loop like I did!

The [documentation] (https://github.com/bitemyapp/bloodhound/blob/master/src/Database/Bloodhound/Client.hs#L683) for `advanceScroll` should be corrected.

bitemyapp:
@MHova I know this is obnoxious and I apologize, but are you sure this isn't true for any of the versions we _do_ support? I don't know if you realized this but we don't officially support >=2.0 right now (breaking tests)

MHova:
Doh, the README says `As of Bloodhound 0.3 all >=1.0 versions of Elasticsearch work.` Well, Bloodhound has been doing fine so far for what I'm doing in ElasticSearch >= 2.0.

Let me see if I can identify exactly which version of ElasticSearch changed this (if at all) and let me update the `advanceScroll` documentation to clearly specify what the terminating condition is for the various versions of ElasticSearch.

I'll also change the README.

bitemyapp:
@MHova much appreciated, thank you. Ping me when I should take another look :+1:

MHova:
@bitemyapp As far as I can tell the terminating condition has always been empty hits even in version 0.9. From the 0.9 [docs](https://www.elastic.co/guide/en/elasticsearch/reference/0.90/search-request-search-type.html):
```
The "breaking" condition out of a scroll is when no hits has been returned.
```

The 1.x docs say the same thing. I updated the PR to add a blurb to the README about 2.0 compatibility, but I left the `advanceScroll` docs the same.

bitemyapp:
@MHova sounds good to me, thank you!

PR #96, opened 2016-01-28
Fix docs specifying an incorrect terminating condition

MHova:
Addresses https://github.com/bitemyapp/bloodhound/issues/95

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,7 +21,7 @@ Endorsements
 Version compatibility
 ---------------------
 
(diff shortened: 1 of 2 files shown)


MHova:
Err....what's up with the CI failure? @bitemyapp

bitemyapp:
@MHova https://travis-ci.org/bitemyapp/bloodhound/jobs/105554260

beats the f out of me. 1.2.4 incompatibility?

MHova:
So what should we do here? I've only changed comments and the README with this PR.

bitemyapp:
@MHova check 1.2.4 docs, see if an API in-compat can be confirmed. Usually if it is, we just drop support for the oldest version and try to add a newer version in exchange, but I don't know that it'd be fair to ask you to do that.

bitemyapp:
@MHova recycling this build to see what it does.

Issue #99, opened 2016-02-02
Support Missing Aggregation


MHova:
Add support for the "Missing Aggregation", which has been available in ElasticSearch since [1.3](https://www.elastic.co/guide/en/elasticsearch/reference/1.3/search-aggregations-bucket-missing-aggregation.html).

MHova:
I'll issue a PR on this.

bitemyapp:
@MHova >available in ElasticSearch since 1.3.

1.2.4 is definitely getting deprecated then :stuck_out_tongue:

MHova:
Ha maybe? 1.2.4 may or may not have actually supported this, but >=1.0 && <1.3 docs are no longer available on the ElasticSearch website.

bitemyapp:
@MHova oh nuts. We should deprecate it anyway if they aren't going to make the bloody docs available. Yeesh.

MHova:
Yeah that's a good idea. For some reason, the 0.9 docs are still available though.

Issue #92, opened 2016-01-23
Restore bloodhound to stackage


DanBurton:
Please release a version of bloodhound that simultaneously supports `aeson-0.9.0.1`, `http-types-0.9`, and `semigroups-0.18.0.1` in order to be restored to Stackage. Even if lts-5.0 is released without bloodhound, we can get it into lts-5.1 or later, as soon as a compatible version is released.

See: https://github.com/fpco/stackage/issues/1155

bitemyapp:
@DanBurton Bloodhound is back on Stackage with Aeson 0.11 right? Can we close this?

DanBurton:
Sure.

Although bloodhound is included in the latest LTS builds, it seems to have been excluded from the nightlies. I've re-enabled it for those as well. https://github.com/fpco/stackage/commit/cfb31f06ad990fdcba78a75d892cac308cdcebf7

bitemyapp:
@DanBurton thank you :)

flip111:
Could V5 be added to stackage?

MichaelXavier:
@flip111 it is. Latest nightly shows version 0.13.0.0 https://www.stackage.org/nightly-2017-03-28/package/bloodhound-0.13.0.0

flip111:
ah ok so just wait for the next LTS 👍

PR #100, opened 2016-02-02
Add Missing Aggregation support

MHova:
Addresses https://github.com/bitemyapp/bloodhound/issues/99

I would write a test for this, but I've been unable to get the tests to run locally.
1) `cabal install -j --only-dependencies --enable-tests` outputs this error:
```
src/Test/Hspec/Core/Type.hs:152:20:
    Constructor ‘HUnitFailure’ should have 2 arguments, but has been given 1
    In the pattern: HUnitFailure err
    In the first argument of ‘E.Handler’, namely
      ‘(\ (HUnitFailure err) -> return (Fail err))’
    In the expression:
      E.Handler (\ (HUnitFailure err) -> return (Fail err))
```

2) To fix the above error, I update the `hspec` bounds in `bloodhound.cabal` from `>= 1.8 && <2.2` to `>= 2.0 && <2.2`

3) Moving to hspec 2.0 causes this error:
```
tests/tests.hs:488:10:
    Duplicate instance declarations:
      instance (Arbitrary k, Ord k, Arbitrary v) => Arbitrary (M.Map k v)
        -- Defined at tests/tests.hs:488:10
      instance [overlap ok] [safe] (Ord k, Arbitrary k, Arbitrary v) =>
                                   Arbitrary (M.Map k v)
        -- Defined in ‘Test.QuickCheck.Arbitrary’
```
I fix it by removing that `Arbitrary` instance.

4) I reinstall and build. When I run `cabal tests` I get the message `You need to re-run the 'configure' command. The version of Cabal being used has changed (was Cabal-1.18.1.3, now Cabal-1.23.0.0).`

5) I run `cabal configure --enable-tests` and re-run `cabal tests`, but it doesn't make a difference.

Edit: Add bit about `Arbitrary`

### DIFF
--- a/src/Database/Bloodhound/Types.hs
+++ b/src/Database/Bloodhound/Types.hs
@@ -237,6 +237,7 @@ module Database.Bloodhound.Types
        , Bucket(..)
        , BucketAggregation(..)
        , TermsAggregation(..)
+       , MissingAggregation(..)
        , ValueCountAggregation(..)
        , FilterAggregation(..)
        , DateHistogramAggregation(..)
@@ -1544,8 +1545,12 @@ data Aggregation = TermsAgg TermsAggregation
                  | DateHistogramAgg DateHistogramAggregation
                  | ValueCountAgg ValueCountAggregation
                  | FilterAgg FilterAggregation
-                 | DateRangeAgg DateRangeAggregation deriving (Eq, Show, Generic, Typeable)
+                 | DateRangeAgg DateRangeAggregation
+                 | MissingAgg MissingAggregation deriving (Eq, Show, Generic, Typeable)
 
+data MissingAggregation = MissingAggregation
+  { maField :: Text
+  } deriving (Eq, Show, Generic, Typeable)
 
 data TermsAggregation = TermsAggregation { term              :: Either Text Text
                                          , termInclude       :: Maybe TermInclusion
@@ -1705,6 +1710,8 @@ instance ToJSON Aggregation where
               , "aggs" .= ags]
   toJSON (DateRangeAgg a) = object [ "date_range" .= a
                                    ]
+  toJSON (MissingAgg (MissingAggregation{..})) =
+    object ["missing" .= object ["field" .= maField]]
 
 instance ToJSON DateRangeAggregation where
   toJSON DateRangeAggregation {..} =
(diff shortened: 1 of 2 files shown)


MichaelXavier:
Have you tried stack? That's how our tests are run in CI.

MHova:
Ah okay thanks. Got it working now after some fiddling. I'll make a test for this PR.

On a completely unrelated note, I'm going to enhance the README with instructions on how to run the tests...

MHova:
@bitemyapp @MichaelXavier Test added and all tests passing. So I guess 1.2.4 does support this kind of aggregation after all?

MichaelXavier:
Cool! Sounds like this is good to merge.

bitemyapp:
@MHova @MichaelXavier thank you, collective of Michael. :busts_in_silhouette:

MHova:
@bitemyapp Thanks! Also, resistance is futile.

Also also, when will this be released on Hackage? I need it for a work project.

bitemyapp:
@MHova I can do a release soon. Most of the time is tracking down all the contributions and changes since last release and writing up the changelog.

MHova:
Thanks for the quick response. I'd like to get this one in the release as well: https://github.com/bitemyapp/bloodhound/pull/89

This one is just a documentation change and is not critical to my project. Hopefully the recycled build will pass: https://github.com/bitemyapp/bloodhound/pull/96

bitemyapp:
@MHova both in.

bitemyapp:
@MHova another reason I was waiting on a release was to decide on a resolution of:  https://github.com/bitemyapp/bloodhound/issues/92

The existing users of the library already went through the pain of migrating to Aeson 0.10 (which probably broke some production code) and we're (I'm speaking for like, 2 people here) not keen to flip back.

Some sort of resolution to this problem is needed, probably requiring mutual Aeson 0.9 and 0.10 compatibility. Open to suggestions.

MHova:
Okay I'll resort to a vendor fork in the meantime then.

bitemyapp:
@MHova well, are you using Aeson 0.10? If so, then that's a vote for doing a release without bothering with LTS compat until things catch up.

MHova:
Yes we are already using aeson 0.10 for our project.

MHova:
Though I'm pretty sure we upgraded from aeson 0.9 to aeson 0.10 for the expressed purpose of using Bloodhound, so we can easily downgrade aeson if necessary.

MichaelXavier:
Same. Bloodhound lead the charge into aeson 0.10 for us too.

bitemyapp:
@MichaelXavier @MHova That's what I suspected. Given that, I am inapt to roll back.

PR #103, opened 2016-02-03
Prepare for internal SumAll Bloodhound build

MHova:
1. Change the library name to sumall-bloodhound
2. Bump major version because the new unreleased content is backwards incompatible
3. Change git repo urls


MHova:
Nope, wrong target.

PR #104, opened 2016-02-03
Doh, need to export BucketValue

MHova:
Now that I've picked up the latest content via our own vendor version...I realize that I forgot to export `BucketValue`! :cry:


bitemyapp:
:smile:

PR #105, opened 2016-02-04
Add helper data types and functions for Missing Aggregations

MHova:
It's funny what you discover once you start eating your own dog food =)

Add the `MissingResult` data type (akin to `TermsResult`) and `toMissing` helper function for parsing `MissingResults` from `AggregationResults`.


MichaelXavier:
@bitemyapp This looks good to me. The more of these we have to do the sillier it will look, since they all have the same definition, but I think having the explicit APIs for parsing these results is better signage than one with a `FromJSON` constraint.

bitemyapp:
@MichaelXavier most of the point is the UX wrapped around the JSON parsing bits anyway. We can try to unify it by some other means later, but I'm fine chucking it in as-is like this for now. Thank you for reviewing this!

bitemyapp:
@MHova and thank _you_ for PR'ing this! :+1: :fireworks: :bear: :bear:

MHova:
:+1: :joy:

PR #106, opened 2016-02-04
Update README

MHova:
Updates for TravisCI ElasticSearch versions. Also remove some unused functions in tests.hs.


bitemyapp:
Thanks again :+1: :smile:

PR #112, opened 2016-03-02
Add list indices feature

MichaelXavier:
Probably the quickest feature add ever for bloodhound. This feature becomes useful if you do time-based indexes and need to retire old data.


bitemyapp:
I think the breakage is my fault, so I'm merging this and focusing on getting this to 0.11.

bitemyapp:
@MichaelXavier thank you as always :)

PR #114, opened 2016-03-16
Aeson lower bound loosening

dmjio:
Since the latest `aeson` isn't on stackage yet, can we loosen the lower bounds of bloodhound to account for all packages `>= 0.8 && < 12`, excluding the problematic `0.10.0.0`

Unless you have an explicit reason to avoid versions `0.8` and `0.9` I don't see why we shouldn't allow them to be used (`0.9` is currently on stackage)

Or, would you want to put `bloodhound` on stackage ? :)

### DIFF
--- a/bloodhound.cabal
+++ b/bloodhound.cabal
@@ -1,5 +1,5 @@
 name:                bloodhound
-version:             0.11.0.0
+version:             0.11.0.1
 synopsis:            ElasticSearch client library for Haskell
 description:         ElasticSearch made awesome for Haskell hackers
 homepage:            https://github.com/bitemyapp/bloodhound
@@ -32,7 +32,7 @@ library
   build-depends:       base             >= 4.3     && <5,
                        bytestring       >= 0.10.0  && <0.11,
                        containers       >= 0.5.0.0 && <0.6,
-                       aeson            >= 0.11.1  && <0.12,
+                       aeson            >= 0.8 && < 0.10 || == 0.11.*
                        http-client      >= 0.3     && <0.5,
                        network-uri      >= 2.6     && <2.7,
                        semigroups       >= 0.15    && <0.19,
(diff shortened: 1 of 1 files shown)


dmjio:
It appears that this will indeed be problematic, since `semigroups` was added to `aeson` and `bloodhound` uses `NonEmpty` :/

dmjio:
I'll have to figure something else out.

bitemyapp:
>Unless you have an explicit reason to avoid versions 0.8 and 0.9 I don't see why we shouldn't allow them to be used (0.9 is currently on stackage)

Not avoiding, just legwork or API incompatibility (via CPP) I'm not keen on.

dmjio:
My only thought would be using `CPP` flags in the cabal file, and then `#ifdef`'ing the `NonEmpty` instances. I assume `aeson-0.11` will be on stackage soon enough though.

bitemyapp:
>then #ifdef'ing the NonEmpty instances.

I don't see how that wouldn't cause there to be a split in the API based on the CPP output.

>I assume aeson-0.11 will be on stackage soon enough though.

Month or two? it's whenever LTS 6.0 hits, but it hasn't mattered since everyone vendors or extra-deps it. Not that I want it to be harder to use - but nobody's complained and I'm not breaking the API just to get the library in LTS sooner.

MichaelXavier:
My strategy has been to throw the advance-copy version of whatever library into extra-deps and at the same time create a calendar event for a week later to check and see if that can be dropped. If not, kick the can down the road another week. Its great if you can use only things in LTS but something not being in LTS shouldn't completely prevent you from moving forward.

One thing worth considering though is that LTS does insist on a GHC version, so people on older versions of GHC are likely going to use extra-deps anyways becuase I don't think stackage makes a habit of backporting new developments to old LTSes that often. It seems like their opinion is to keep the Haskell community in sync with the cadence of GHC releases.

PR #117, opened 2016-04-17
Fix Haddocks italicizing forward slashes

MaxGabriel:
The haddocks are currently reading the forward slashes to mean to italicize:

image


bitemyapp:
ahhhh, thank you!

PR #119, opened 2016-05-26
add sniffing

winterland1989:
It's a quite easy change since we have a central `dispatch` already, the main issue here is that i changed the way how to construct `Server`, so a major bump is neccessary. I also cleaned up the unnecessary `bindM2`s. please feedback : )


winterland1989:
The travis fail is a little mystery to me, any clue?

bitemyapp:
@winterland1989 the CI tests across multiple ES versions, in this case ES 1.2.4 failed, the test case is here: https://travis-ci.org/bitemyapp/bloodhound/jobs/133057705

winterland1989:
It's fixed in upstream, merged ;)

bitemyapp:
@winterland1989 I'm not comfortable with this being in the README:

```haskell
testServer <- newServerNoSniff "http://localhost:9200"
```

The default path for things needs to work for most people and to be properly tested. Taking your code out of the test suite means I am unlikely to merge it.

I would rather discuss deprecating ES 1.2.4 than make the sniffing code go untested. Ideally with some data on what ES versions are in popular use, if possible.

I'd like @MichaelXavier to pipe up on this as well before moving to a merge.

winterland1989:
+1 on adding sniffing test, but it's really trick to test this without a dynamic changing cluster, and the test process can be very time consuming, so please go on discussing on this, i need more input too, while i'm testing my sniffing code on my company already.

There're more design question to be discussed too, for example. how to deal with sniffing errors?

I think this is a must-have feature for any serious es usage, if all request go through master node, then deploying a cluster is pointless. so i'd like to see it in next major release.

bitemyapp:
@winterland1989 I just merged @MichaelXavier's PR which deprecated 1.2.4, would you be open to catching this up to the current state of master and fixing up the tests/UX/docs situation for this functionality?

To reiterate what I said earlier, I want it to minimally perturb people that are already using it. It should not silently change the behavior of existing functions. If that's alright, I'd like to see this wrapped before I do another release, if possible.

I should probably add GHC 8.0 support soon as well.

MichaelXavier:
Sorry for never commenting on it. I didn't know about this sniffing thing, I just use an http load balancer. I'll check it out. I also [PRed](https://github.com/winterland1989/roundRobin/pull/1) `roundRobin` to use a `NonEmpty` for its API so it isn't partial. This brings it in line with our coding standards in bloodhound.

winterland1989:
I'll try to get back to this next week, thanks!

bitemyapp:
@winterland1989 we've got someone pushing on 2.0 compat which might churn things for you, I'm going to revisit your PR after that's done unless you can wrap it up soon.

winterland1989:
That's fine, I'm just too busy with my job these days.

PR #120, opened 2016-05-31
Make it possible to not specify id so it's generated by ElasticSearch

arybczak:
There seems to be no possibility to insert a new document without specifying an id and letting ElasticSearch generate it, so this pull request provides this functionality.

I think I didn't properly update documentation, but I would like to confirm first that there is indeed no way to do what I described and that this pull request is useful and will be merged if documentation is fixed.


bitemyapp:
I do not recommend letting Elasticsearch generate your document IDs if you want to keep your data and I'm unlikely to break the API to enable doing so. I recommend using UUIDv4 or Flake-style IDs instead.

I can be argued from this position, but I don't think arguments from convenience will suffice to get this PR in.

MichaelXavier:
Throwing out an idea, it may be stupid, but what if we provided a `Random` instance for `DocId` that internally used the `uuid` package and converted it to Text. This is probably what most folks are doing anyways, but it would provide a reasonable default and seems pretty opt-in. One downside would be additional transitive dependencies from `uuid`. I'm not sure if the current thinking encourages UUID for elasticsearch ids or not, but its seemingly worked out well for me so far.

arybczak:
> I do not recommend letting Elasticsearch generate your document IDs if you want to keep your data

Why? I am never going to use IDs to search for the data and the data itself will not persist for a very long time (log messages in this particular case), so I shouldn't have to care about them at all.

> I recommend using UUIDv4 or Flake-style IDs instead.

From ElasticSearch documentation: `Autogenerated IDs are 20 character long, URL-safe, Base64-encoded GUID strings. These GUIDs are generated from a modified FlakeID scheme which allows multiple nodes to be generating unique IDs in parallel with essentially zero chance of collision.` Why would I want to do the same thing myself?

An optional way to solve it is to treat empty DocId as a special value and let ES generate it in such case. This solves API breakage, but changes the behavior if you pass empty id and is arguably unidiomatic Haskell, that's why I chose to provide a version with Maybe.

A fundamental problem that results from bloodhound not allowing you to make ES generate IDs is that not only it forces you to care about things you possibly do not wish to care about, but also encourages suboptimal design decisions. As an example, consider katip-elasticsearch library. They generate IDs in the following way:
```haskell
mkDocId :: IO DocId
mkDocId = (DocId . T.decodeUtf8 . toASCIIBytes) `fmap` randomIO
```
This is not particularly good solution and if they weren't forced to pass them, then they wouldn't care and ES would generate IDs of much better quality, so everybody would win.

@MichaelXavier I personally do not agree with duplication of functionalities in the client library if a server can do the same thing just fine.

bitemyapp:
@arybczak >a modified FlakeID scheme

Wasn't always true. They used to use auto-incrementing ids that could collide, and upon a cluster rejoin after a split-brain, could lose data. If they've fixed it, I'm willing to consider relaxing this constraint, but I'd like to have a chance to examine their id generation scheme in a little more detail.

arybczak:
From documentation of 1.x: `Autogenerated IDs are 20 character long, URL-safe, Base64-encoded string universally unique identifiers, or UUIDs.`.

I don't see why you would want to deprecate older version as auto generation will be optional.

bitemyapp:
@arybczak because most people don't realize the terrifying ways in which you can lose data in ES. Or that they shouldn't use it as a ground source of truth for their data. It's damage control.

I've bulk inserted large datasets into an ES cluster in the past and had thousands of documents just…missing. Not the indexer catching up — just gone. You just get used to this sort of thing with ES and assuming that you have to validate that after ES says you inserted something that it's actually there and in the correct state. Part of the problem is that these problems don't really surface until you start using it as an actual cluster with a non-trivial dataset or workload. They don't happen when you're putzing around with 1,000 documents on a single node cluster running on your macbook pro.

Pushing back on this inclination to avoid leading people to doing things that could cause them to lose data does not get your PR in faster.

MichaelXavier:
Should we close this PR then?

bitemyapp:
@MichaelXavier yep, thanks.

PR #123, opened 2016-07-08
Support replica count of 0

MichaelXavier:
I can see why the smart constructor for shard count has a minimum of 0
but a replica count of 0 is totally valid.


bitemyapp:
@MichaelXavier wat

PR #125, opened 2016-07-21
Set minimum lower bound on http-client

MichaelXavier:
There were either some functions removed in the http-client 0.4 line or
ones added in 0.5 but after binary searching the version range, 0.4.30
seems to be the magic number where everything compiles and the tests
pass without having to do any CPP or worse.


bitemyapp:
danke danke :dancers:

PR #127, opened 2016-08-08
Break up client API into sections

MichaelXavier:
Just a simple change. Client functions already roughly followed a grouping, but I added headings to the client so the haddocks are less daunting. It would probably be a good idea to do this some day to the types module but this is a good start.


bitemyapp:
@MichaelXavier a nice touch, thank you :)

PR #130, opened 2016-08-21
Added the OverloadedStrings extension to first example in README.md

dinnu93:
The first example doesn't work without the OverloadedStrings extension. The Server keeps demanding it's argument to be of the type Text


bitemyapp:
ty ty :+1:

PR #133, opened 2016-08-24
Fix parent child tests for 2.0.

alistair:
So it seems that to create parent child mappings in 2+ you must first create the child mapping AND then the parent.

Also 2+ has made indexes more strict.

If an index has multiple types which have matching properties then the types of those properties must be the same.

So in this example both the TweetMapping and ParentMapping must have the same types for the user, message, age, etc properties

maybe better explained by https://www.elastic.co/guide/en/elasticsearch/reference/2.0/breaking_20_mapping_changes.html#_conflicting_field_mappings

### DIFF
--- a/tests/tests.hs
+++ b/tests/tests.hs
@@ -174,15 +174,32 @@ data ParentMapping = ParentMapping deriving (Eq, Show)
 
 instance ToJSON ParentMapping where
   toJSON ParentMapping =
-    object ["parent" .= Null ]
+    object ["parent" .=
+              object ["properties" .=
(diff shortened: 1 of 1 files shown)


bitemyapp:
@alistair needs guarded for 1.4 or above?

alistair:
Mmmm.... interesting,  No sure I see how its related to anything I changed.  Will investigate over the weekend probably.

alistair:
FYI I just pushed an empty commit to trigger travis again.   I would really like to rebase that out of the pull request first.

alistair:
Yeah as I thought.  It was a transient failure.

PR #135, opened 2016-08-30
Format date as dd-MM-yyyy to support stricter 2.0+

alistair:
So this lets an additional test pass on ES 2.0 due to stricter date parsing ( no more than that sadly )

It changes the date formatting to have leading 0's e.g.  02-02-2002.

It also removes more code than it adds :)


bitemyapp:
Seems like a straight-forward win to me. Thank you :)

PR #137, opened 2016-09-15
Node stats api

MichaelXavier:
I just needed 1 field of this but the completionist (or masochist?) in me decided to implement the parsing of the whole stats API. As with the last type of API here, I pretty much can only use the tests as a smoke test since the stats are highly variable, but they are passing. I've merged in master so I should be up to date there.

### DIFF
--- a/src/Database/Bloodhound/Types.hs
+++ b/src/Database/Bloodhound/Types.hs
@@ -1739,7 +1739,7 @@ data TermsAggregation = TermsAggregation { term              :: Either Text Text
                                          , termAggs          :: Maybe Aggregations
                                     } deriving (Eq, Read, Show, Generic, Typeable)
 
-data CardinalityAggregation = CardinalityAggregation { cardinalityField :: FieldName,
+data CardinalityAggregation = CardinalityAggregation { cardinalityField   :: FieldName,
                                                        precisionThreshold :: Maybe Int
                                                      } deriving (Eq, Read, Show, Generic, Typeable)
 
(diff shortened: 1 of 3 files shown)


src/Database/Bloodhound/Types.hs
```diff
@@ -4373,6 +4667,356 @@ instance FromJSON NodesInfo where
...
+        warmer <- o .: "warmer"
+        flush <- o .: "flush"
+        refresh <- o .: "refresh"
+        merges <- o .: "merges"
+        search <- o .: "search"
+        getStats <- o .: "get"
+        indexing <- o .: "indexing"
+        store <- o .: "store"
+        docs <- o .: "docs"
+        NodeIndicesStats <$> (unMS <$> recovery .: "throttle_time_in_millis")
```
bitemyapp: Code like this makes me think we want `lens-aeson`.

bitemyapp:
@MichaelXavier Looks good to go except for the test failures.

src/Database/Bloodhound/Types.hs
MichaelXavier: Yeah. And it gets hairier since I'm finding from these test failures that the higher level keys like "recovery" are optional, so I'm defining a little scoped combinator `(.::)` That works where the object is a Maybe. You know how folks get about lenses though.

MichaelXavier:
This one's a new one, even on the same version of ES the available keys change. I'm running the parser against some of the reported bodies from travis and making things Maybe where necessary.

bitemyapp:
/me shakes head and grumbles about Elasticsearch

MichaelXavier:
We're green!

Issue #138, opened 2016-09-22
`nodePluginVersion` field in `NodePluginInfo` needs to be a 'Maybe VersionNumber'


bermanjosh:
I was failing tests for `getNodesInfo` and `getNodesStats`. Turns out some elastic plugins return "NA" as their version (e.g. "head" plugin).

bermanjosh:
Similar issue with `nodeOSLoad` field of `NodeOSStats` - Windows seems to not expose this info. I'm making it a `Maybe` instead of a CPP.

(I had to do a little linux / windows work in general, but I'm not committing that as part of this).

bermanjosh:
Also - thanks to @ygale for his help

PR #139, opened 2016-09-22
Wrap VersionNumber and LoadAvgs in Maybe #138

bermanjosh:
The version number issue is OS-agnostic, the LoadAvgs issue is Windows-specific. 

The latter could have been dealt with using a CPP, but I generally try to avoid that. 

This is a breaking change, but it's recent stuff, so I didn't know what you wanted me to do with the version numbers. (Also, I had to make other changes there for windows, which I'm going to keep separate until my 2.4 work pushes forward).

### DIFF
--- a/src/Database/Bloodhound/Types.hs
+++ b/src/Database/Bloodhound/Types.hs
@@ -4064,7 +4064,7 @@ data NodeOSStats = NodeOSStats {
     , nodeOSCPUIdle        :: Int
     , nodeOSCPUUser        :: Int
     , nodeOSCPUSys         :: Int
-    , nodeOSLoad           :: LoadAvgs
+    , nodeOSLoad           :: Maybe LoadAvgs
     , nodeOSUptime         :: NominalDiffTime
     , nodeOSTimestamp      :: UTCTime
     } deriving (Eq, Show, Generic, Typeable)
@@ -4189,7 +4189,7 @@ data NodePluginInfo = NodePluginInfo {
     , nodePluginJVM         :: Bool
     -- ^ Is this plugin running on the JVM
     , nodePluginDescription :: Text
-    , nodePluginVersion     :: VersionNumber
+    , nodePluginVersion     :: Maybe VersionNumber
     , nodePluginName        :: PluginName
     } deriving (Eq, Show, Generic, Typeable)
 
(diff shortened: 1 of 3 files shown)


bermanjosh:
Also - thanks to @ygale for his help

bermanjosh:
@MichaelXavier 
How's this? It passes all the tests. I got caught in a silly loop in an earlier version, so give it a once-over.

src/Database/Bloodhound/Types.hs
```diff
@@ -5212,3 +5213,10 @@ instance FromJSON NodeNetworkInterface where
       parse o = NodeNetworkInterface <$> o .: "mac_address"
                                      <*> o .: "name"
                                      <*> o .: "address"
+
+newtype MaybeNA a = MaybeNA { unMaybeNA :: Maybe a }
+  deriving (Show, Eq)
+
+instance FromJSON a => FromJSON (MaybeNA a) where
+  parseJSON (String "NA") = pure $ MaybeNA Nothing
+  parseJSON o = MaybeNA . Just <$> parseJSON o
```
bitemyapp: Not my preference, destroys information.

Prefer "NA" be an explicit possibility in a sum type for versions. /cc @MichaelXavier thoughts?

MichaelXavier: I recommended it be done this way. To me, NA is semantically meant to indicate that the version is not available. It would make sense to me if I saw it in the haddocks, plus I get to use all of my tools such as `maybe` and the `Maybe` monad to work with the value still whereas all I've got is a `case` if we added a constructor for NA version.

bitemyapp: That's fine then. I don't like the loss of provenance, but if it suits you, I won't object.

bermanjosh: @bitemyapp 
Ok, so any other changes you want? What do you want me to do with the version number?

bitemyapp: @bermanjosh I think the conclusion of @MichaelXavier's and I back and forth was "no changes so far"

MichaelXavier:
So I'm ok to merge this if you are, @bitemyapp.

bitemyapp:
One request, so I don't have to keep plumbing the git log for this: @bermanjosh could you add yourself and what you did to the changelog? Once that's in, I'll merge it.

bermanjosh:
@bitemyapp OK, then should I bump the version? 0.11.1.0?

bitemyapp:
http://semver.org/

So far as I can tell, your changes fall under `MAJOR`. Usually we try to wait and roll up more changes before cutting a `MAJOR`. This fits the scenario since you have prod testing to do for future changes as well.

bermanjosh:
Ok, so then I don't think it makes sense to do this now; unless you're ok with me adding to the top of the changelog without a version.

bitemyapp:
@bermanjosh hwat

bermanjosh:
Ah, got it.

bermanjosh:
@bitemyapp done

bitemyapp:
@bermanjosh tyvm!

PR #142, opened 2016-09-29
Comment out comments in the example properly.

23Skidoo:


bitemyapp:
@23Skidoo what sad turn of fate has you looking at an Elasticsearch client?

23Skidoo:
@bitemyapp We're using it at $JOB.

bitemyapp:
@23Skidoo my condolences

23Skidoo:
@bitemyapp It's only for logging, so I hope to preserve my sanity.

bitemyapp:
@23Skidoo That should be okay actually! I'd still broadcast the logs onto the local disk and maybe somewhere else if they're important (Could vary this behavior by severity)

YMMV.

https://github.com/Soostone/katip has some ways of doing this if you haven't seen it.

23Skidoo:
@bitemyapp Thanks!

PR #143, opened 2016-10-03
Fix an example.

23Skidoo:
No need to specify the mapping name twice, since `putMapping` does `PUT /$INDEXNAME/_mappings/$MAPPINGNAME` and you give it an `IndexName` and a `MappingName`. In fact, specifying the mapping name here breaks upsert semantics (you get `Root type mapping not empty after parsing!` exception when trying to add the same mapping twice).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,9 +1,11 @@
 cabal.sandbox.config
 .cabal-sandbox/
 dist/
+dist-newstyle/
 .project
 .dist-buildwrapper/
 .DS_Store
+TAGS
 *.hi
 *.o
 bloodhound-*-docs/
--- a/README.md
+++ b/README.md
@@ -181,10 +181,9 @@ data TweetMapping = TweetMapping deriving (Eq, Show)
 :{
 instance ToJSON TweetMapping where
   toJSON TweetMapping =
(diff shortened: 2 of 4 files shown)


23Skidoo:
I test with ES 1.7, so this was broken in ES < 2 already, despite what [documentation says](https://www.elastic.co/guide/en/elasticsearch/reference/1.7/indices-put-mapping.html).

23Skidoo:
Also updated tests and made some other small fixes.

23Skidoo:
Interesting, the `UpdatableIndexSetting` roundtrip property seems to have been broken by the 

```
instance ApproxEq Char where
+  (=~) = (==)
```

change.

bitemyapp:
>Not sure how to fix.

/cc @MichaelXavier Didn't you need `ApproxEq` for something?

23Skidoo:
Couldn't reproduce the failure locally with `./dist/build/tests/tests --seed 47535321`.

23Skidoo:
@bitemyapp I backed out the GHC 8 compatibility fix, so this should be now good to go.

bitemyapp:
@23Skidoo thank you!

PR #145, opened 2016-10-18
Precise return type for the `bulk` operation

bartavelle:
This patches introduces the following changes :
- Add a description field to the `EsProtocolException` constructor. That was useful when checking the aeson code, and I think it would be useful when users report errors if the ES server protocol changes. This might be part of a distinct patch/PR though.
- Several new types related to bulk operations responses are introduced.
- The `bulk` function has been altered. It now throws an `EsProtocolException` when the json can't be decoded.
- A few tests have been added that should exercise the json content.

It works for me (tm), but naming could definitely be improved. I will be away for a few days, but should be able to integrate all remarks by the end of next week.

### DIFF
--- a/src/Database/Bloodhound/Client.hs
+++ b/src/Database/Bloodhound/Client.hs
@@ -613,8 +613,8 @@ parseEsResponse reply
         tryParseError = case eitherDecode body of
                           Right e -> return (Left e)
                           -- this case should not be possible
-                          Left _ -> explode
-        explode = throwM (EsProtocolException body)
+                          Left errdesc -> explode errdesc
+        explode rr = throwM (EsProtocolException body rr)
 
 -- | 'indexExists' enables you to check if an index exists. Returns 'Bool'
 --   in IO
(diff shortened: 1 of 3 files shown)


bitemyapp:
Restarting CI job

MichaelXavier:
Do we know what happened with the build? Master is building fine but the whole matrix seems broken in this PR, even after restarting CI.

bitemyapp:
@MichaelXavier I'm not sure, @bartavelle would need to debug and dump the requests. It seems to mostly be a lot of this,

screenshot from 2016-11-22 15-27-47

My _guess_ is that the request body is somehow malformed as a result of the changes, but absent time spent investigating, I cannot say for sure.

bartavelle:
I didn't realize things broke ... will check again.

bartavelle:
However, this PR didn't seem to attract much interest, is that worth it I spend my tome on that ?

bitemyapp:
@bartavelle What version of Elasticsearch did you test this with?

bartavelle:
2.4.1. I installed it just before writing the PR, but it seems it's now version 5.0.1, so perhaps I did something really wrong and downloaded a very old version (?).

bitemyapp:
@bartavelle and all the tests passed on 2.4.1?

Bloodhound doesn't officially support 2.x at the moment, let alone 5.x.

https://github.com/bitemyapp/bloodhound#version-compatibility

Some people use it with those versions anyway, but until we figure out how to handle the constant breakage Elasticsearch subjects the API to, it stays put. Most commercial users of Elasticsearch don't seem to notice or track the constant updates, so it's unlikely we'd do that either given the labor constraints.

bartavelle:
Many tests broke, but mine worked ... but I suppose it's the other way around with the versions that are supported! I'll try to find the time to see how it works for the other versions.

bitemyapp:
This has a merge conflict and hasn't been touched in awhile, closing.

Issue #146, opened 2016-10-28
Stop using deprecated doctest-prop


DanBurton:
doctest-prop has been marked as deprecated.

bitemyapp:
Pinging author of #147 for this.

bitemyapp:
Fixed by #147's merge.

PR #147, opened 2016-10-31
Drop doctest-prop

MichaelXavier:
AFAICT we weren't even using it anymore. We imported the function assert
and then never used it in the doctests.

This is for #146. We'll see what travis says.

### DIFF
--- a/bloodhound.cabal
+++ b/bloodhound.cabal
@@ -92,7 +92,6 @@ test-suite doctests
                       bloodhound,
                       directory,
                       doctest >= 0.10.1,
-                      doctest-prop,
                       filepath
   else
     buildable: False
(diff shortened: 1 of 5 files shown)


bitemyapp:
@MichaelXavier This test failed,

`  1) JSON instances UpdatableIndexSetting FromJSON/ToJSON roundtrips`

Any ideas?

MichaelXavier:
That's unrelated to the fix here but I was meaning to look into it. There has to be an ambiguity in the json parsers that got exposed by chance thanks to QuickCheck. As you can see, the tests were actually run twice on travis and passed once.

Since its unrelated, its fine to merge with that broken. I've created #149 and self-assigned to fix the quickcheck error. I imagine if we threw a lot of CPU time at the problem (by setting quickcheck minimum tests really high) we could catch a couple more of these. Wouldn't necessarily want to do that in travis as it would slow the tests way down.

bitemyapp:
@MichaelXavier dunzo, thank you!

PR #150, opened 2016-11-03
Fix quickcheck failure due to ambiguous parse

### LINKED ISSUE #149: Quickcheck fails with seed 1360169055
Exposed by https://travis-ci.org/bitemyapp/bloodhound/jobs/172125367

I'll look into it.

MichaelXavier:
Due to the way node attr filters are serialized (attr name as an object
key), its possible to express a value in haskell that will parse
ambiguously. If we wanted to move the invariant into the types we could
use a map I guess but whatever.

I ran this updated test through 100,000 tests and we're all green.

This resolves #149

### DIFF
--- a/tests/tests.hs
+++ b/tests/tests.hs
@@ -1,5 +1,6 @@
 {-# LANGUAGE CPP                        #-}
 {-# LANGUAGE DefaultSignatures          #-}
+{-# LANGUAGE DeriveDataTypeable         #-}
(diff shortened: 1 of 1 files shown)


MichaelXavier:
@bitemyapp :+1:

bitemyapp:
@MichaelXavier ty ty :smile_cat:

PR #151, opened 2016-11-03
add function for checking index status

andrewthad:
This is helpful for writing integration tests that tear down an index and then immidiately try writing to it.


bitemyapp:
Schwifty!

bitemyapp:
You can say "Fixes #152" in future so that merging auto-closes the issue btw

andrewthad:
Ah, cool trick. I'll try that out next time.

PR #153, opened 2016-11-17
Relax aeson and hspec upper bounds

phadej:
Also update stack-8.0.yaml to use newer nightly, with aeson-1.0.*


bitemyapp:
Did you test the Aeson bump? That has bitten us before.

phadej:
I trust your tests.

There aren't any breaking changes in `aeson-0.11` → `aeson-1.0`, which wouldn't result in type error. At least we (bergmark, and winterland1989) paid attention to that. If that's not true, it's a bug in `aeson` and we will fix it.

bitemyapp:
Good enough for me, thanks!

PR #155, opened 2016-11-18
Switch derive to generics-sop

phadej:
Until https://github.com/nick8325/quickcheck/pull/112 is merged and
released

AFAICS, this is the last bit, so we can put `bloodhound` back on Stackage again.

### DIFF
--- a/tests/tests.hs
+++ b/tests/tests.hs
@@ -7,9 +7,11 @@
 {-# LANGUAGE OverloadedStrings          #-}
 {-# LANGUAGE RecordWildCards            #-}
 {-# LANGUAGE ScopedTypeVariables        #-}
-{-# LANGUAGE TemplateHaskell            #-}
 {-# LANGUAGE TypeOperators              #-}
 {-# OPTIONS_GHC -fno-warn-orphans #-}
+#if __GLASGOW_HASKELL__ < 800
+{-# OPTIONS_GHC -fcontext-stack=100 #-}
+#endif
 module Main where
 
 import           Control.Applicative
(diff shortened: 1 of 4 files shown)


tests/tests.hs
```diff
@@ -577,7 +578,8 @@ instance ApproxEq BoolMatch
 instance ApproxEq MultiMatchQuery
 instance ApproxEq IndexSettings
 instance ApproxEq AllocationPolicy
-instance ApproxEq Char
+instance ApproxEq Char where
```
phadej: my compiler said there is no `Generic Char` instance.

tests/tests.hs
```diff
@@ -794,114 +796,114 @@ instance Arbitrary VersionNumber where
...
-$(derive makeArbitrary ''IndexSettings)
-$(derive makeArbitrary ''UpdatableIndexSetting)
-$(derive makeArbitrary ''Bytes)
-$(derive makeArbitrary ''AllocationPolicy)
-$(derive makeArbitrary ''InitialShardCount)
-$(derive makeArbitrary ''FSType)
-$(derive makeArbitrary ''CompoundFormat)
-$(derive makeArbitrary ''FsSnapshotRepo)
-$(derive makeArbitrary ''SnapshotRepoName)
+instance Arbitrary IndexName where arbitrary = sopArbitrary; shrink = genericShrink
```
phadej: When `QuickCheck` gets generic derivation, we'll be able to drop the `where` statements

tests/tests.hs
RyanGlScott: That's by design ;) https://ghc.haskell.org/trac/ghc/ticket/9526#comment:13

MichaelXavier:
Thanks for this! I'm not holding my breath on that PR escaping the shadow realm any time soon. I'll be implementing this on my own packages that are depending on derive for arbitrary instances.

MichaelXavier:
@bitemyapp Any objection to merging this? According to [stackage](https://github.com/fpco/stackage/blob/master/build-constraints.yaml#L519) bloodhound is disabled from nightlies due to derive and "aeson 1.0". I assume that means it didn't support aeson 1.0 but now its < 1.1 so we should be good there. I'm on a tear going through all my packages and getting them back on stackage and katip-elasticsearch I think is blocked on bloodhound. I move that we move for a release, so I'll look at the other outstanding PRs, then we can write up a changelog and release!

bitemyapp:
@MichaelXavier yep, I'll merge this now. Just keep pinging if you need anything or think it's time to cut the release.

MichaelXavier:
I doubt the last standing PR is going to get resolved soon, so I will go ahead and review the diff between last released version and master, choose an appropriate version bump, update the changelog and submit a PR for release.

bitemyapp:
@MichaelXavier sounds good, thank you!

Issue #157, opened 2016-11-29
Elastic 5.0 support


bermanjosh:
Hey,

I've been working on getting bloodhound to work against Elastic 5.0, here:
https://github.com/bermanjosh/bloodhound/tree/es5

It's not a fully-featured fork, but I've got I think I've got the major semantic changes (such as Filters), and all the tests pass. I've been slowly incorporating this into my application, and I've got a working POC (some testing is still outstanding, though).

I know in the past we've discussed trying to keep compatibility and sharing modules, but wasn't able to find a clean way to get that to work. I'm opening this issue for you guys to see what I've done, and decide if it's something you'd like to try to work back in. 

Please let me know what you think.

bitemyapp:
>but wasn't able to find a clean way to get that to work.

Yeah this doesn't surprise me at all. One way forward is to make the repo dual package.

Let us take the temperature of things. @MichaelXavier - what do you think? Where's Amazon's ES support at? I think they support 1.5 and 2.3 at the moment: https://aws.amazon.com/elasticsearch-service/faqs/

So, for decent coverage we'd need a bloodhound-1, bloodhound-2, bloodhound-(5|current).

@bermanjosh the commit that your branch is ahead on, is that all that was needed to bring bloodhound up to compatibility with 5.x? All tests passing?

bermanjosh:
@bitemyapp
All tests pass, yes.

I squashed everything into one commit. 
https://github.com/bermanjosh/bloodhound/commit/437bc3d16f00bb2de339a682e3df866d0518e94d

MichaelXavier:
I'm no longer using the Amazon ES service for the typical reasons you'd expect with any new-ish AWS product launch. I'm running my own cluster on 1.7. However, I'd say the main reason why I'm on 1 and not 2 is due to bloodhound not officially supporting 2, so its kind of a catch 22. FWIW (small sample size) most people I talk to (non Haskellers, naturally) are on ES 2 or 5.

I'd be willing to go to 2 for now. I have no objection to going to 5 either but I'm pretty sure you want to do major version upgrades 1 at a time. I'm not sure how well either the fork or master works on 2, but I imagine any upgrade would result in downtime: you'd bring the app down, snapshot, restore to the new cluster, bring the new version of the app online with the newer bloodhound. I think that's going to be the hard sell but its gotta happen eventually.

[...]

PR #158, opened 2016-12-01
Es5

bermanjosh:
PR for #157


MichaelXavier:
I'd recommend updating the test matrix to at least include ES 2.4.2 for one of the GHC versions. Probably would be good to start with 1 until its passing.

.travis.yml
```diff
@@ -4,19 +4,19 @@ addons:
   apt:
     packages:
       - libgmp-dev
-      - openjdk-6-jre
+      - oracle-java8-installer
```
bitemyapp: Was this to diagnose/eliminate a runtime problem? I'd like to understand the motivation.

bermanjosh: Elastic 5.x needs Java 8 [1]
Travis only supports Oracle JDK 8 [2]

[1] - https://www.elastic.co/guide/en/elasticsearch/reference/current/setup.html#jvm-version
[2] - https://docs.travis-ci.com/user/ci-environment/#JDK

bitemyapp: Thanks!

bermanjosh:
@bitemyapp @MichaelXavier 
Hey - the merge is clean and tests pass. What's next?

bitemyapp:
Do we block on explicit ES2 compat?

Also I'll be flying out to Colorado for the LC Winter Retreat, so my availability is going to be spotty for the next week.

bermanjosh:
Personally, I don't think it's worth maintaining. If no one was clamoring for it to this point, and we've now eclipsed it, I don't see a reason to develop for it. 

I can run the V5 modules against ES2 locally and see if / how badly they fail. We can then add some appropriate note to the readme based on that.

bermanjosh:
@bitemyapp Oh, also - enjoy! GL on the talk.

bitemyapp:
I'm fine lazily evaluating on ES2 support but somehow thought it was something that would be nice for @MichaelXavier - I may be mis-remembering.

bermanjosh:
I thought @MichaelXavier meant just to see. 

My concern is if it fails (which isn't totally unlikely)? If we're not committing to fixing those failures, won't that fail every subsequent commit?

bitemyapp:
>I thought @MichaelXavier meant just to see.

That tickles a part of my memory yes.

>My concern is if it fails (which isn't totally unlikely)? If we're not committing to fixing those failures, won't that fail every subsequent commit?

We won't block on that then.

MichaelXavier:
Sorry for the silence. I'm on 1.7 right now and I think the official path is 1 -> 2 -> 5. However, I'm not going to tax your efforts @bermanjosh. My time on 2 will probably be brief. So I say don't let 2 hold up 5. If it works for 2: great. If it doesn't and I need it for upgrade, 5 will serve as a model for 2 and I may contribute 2 support back if I end up needing it.

bermanjosh:
@MichaelXavier Ok, thanks. 

I just pushed an update to include tests for 2 against both the V1 and V5 modules. When I tested them locally, I got 14 and 15 failures. So, basically, the situation hasn't really changed. Es2 might work, but nothing's guaranteed.

MichaelXavier:
We may want to disable that part of the test matrix before it goes to master so we have a green suite there.

bermanjosh:
I think I did. Though it looks like it's hiding the fact I put the links in incorrectly.

.travis.yml
```diff
@@ -4,19 +4,26 @@ addons:
...
+   - GHCVER=7.10 ESVER=1.5.2 ESFLAG=ES1 DLINK=https://download.elastic.co/elasticsearch/elasticsearch/elasticsearch
+   - GHCVER=7.10 ESVER=1.6.0 ESFLAG=ES1 DLINK=https://download.elastic.co/elasticsearch/elasticsearch/elasticsearch
+   - GHCVER=7.10 ESVER=1.7.2 ESFLAG=ES1 DLINK=https://download.elastic.co/elasticsearch/elasticsearch/elasticsearch
+   - GHCVER=8.0  ESVER=1.7.2 ESFLAG=ES1 DLINK=https://download.elastic.co/elasticsearch/elasticsearch/elasticsearch
+   - GHCVER=8.0  ESVER=2.4.2 ESFLAG=ES1 DLINK=https://download.elastic.co/elasticsearch/release/org/elasticsearch/distribution/tar/elasticsearch/2.4.2/elasticsearch
+   - GHCVER=8.0  ESVER=2.4.2 ESFLAG=ES5 DLINK=https://download.elastic.co/elasticsearch/release/org/elasticsearch/distribution/tar/elasticsearch/2.4.2/elasticsearch
+   - GHCVER=8.0  ESVER=5.0.2 ESFLAG=ES5 DLINK=https://artifacts.elastic.co/downloads/elasticsearch/elasticsearch ES_JAVA_OPTS="-Xms500m -Xmx750m"
+
+matrix:
+  allow_failures:
```
bermanjosh: @bitemyapp @MichaelXavier 
Any idea why this isn't allowing the es2 failures through?

MichaelXavier: Going by the docs this seems like this should work. I assume its just travis not working as indicated.

bermanjosh: So should I take the tests out?

MichaelXavier: I'd say sure, or comment them out. I imagine if I'm working on v2 backports I'll create a branch, reenable the tests and work from there.

bermanjosh: @MichaelXavier

bermanjosh:
@bitemyapp I'll try to bring in @MichaelXavier's aeson work over the weekend, or early next week.

bermanjosh:
@bitemyapp What's next?

bitemyapp:
@bermanjosh @MichaelXavier merged so we can just make a mess / fix things in master. Next release should include the split modules. As I understand it, we're only committing to V1/V5 for now unless someone steps up for V2.

/cc @MHova

Issue #162, opened 2017-01-02
Aeson 1.1 compatibility


bitemyapp:
https://github.com/fpco/stackage/issues/2177

Ahoy mateys.

/cc @bermanjosh @MichaelXavier 

Mind if I work towards compat on this? https://github.com/bitemyapp/bloodhound/pull/158 still has merge conflicts, should I push forward separately?

bermanjosh:
Yeah, of course not. I'll rebase when you're done - but once we're on it. Are those conflicts on #158 things you want me working on? Sorry if I misunderstood, I thought that was stuff that needed to be done someone with write access.

bitemyapp:
@bermanjosh  I'm not sure I understand. It's your PR from your fork on your es5 branch. You can reconcile it with master on your local copy, push to your fork's es5 branch, and the PR will update. Have I missed something?

It sounds like I have the thumbs up from you to push forward with a release for Aeson 1.1 compat otherwise? I wanted feedback because Aeson is a touchy one.

bermanjosh:
@bitemyapp No, you didn't miss anything, **I** did. Sorry. I'll do that tomorrow. 

As for the Aeson, yeah, go for it, don't let me hold you back. I'll take whatever you do and push it into my fork as well when you're done.

bitemyapp:
Well, it's not just about your fork. I think Aeson will avoid breakage going forward but I'm a bit touchy about updating it after the `0.10` debacle.

bermanjosh:
@bitemyapp In theory I'd say better to wait a week and see if anything evil happens, but I don't think it's worth falling behind. What's the time limit for stackage?

MichaelXavier:
Anything I can help on with this? I'm getting pinged via the aeson1.1 mega ticket [here](https://github.com/fpco/stackage/issues/2177) that several of my libraries need an upgrade. Bloodhound dep in katip-elasticsearch is tripping me up. The changelog for 1.1 looks pretty innocuous.

bitemyapp:
@MichaelXavier still at conf and won't be home and able to work until like Saturday. If you'd like it done before then, I certainly wouldn't discourage you cutting a release yourself before we resolve the new split modules stuff.

Otherwise, I have to imagine I'll have time for it in the coming weekend.

MichaelXavier:
And we're green! Should I go ahead and go from `0.12.0.0` -> `0.12.1.0` and publish?

bitemyapp:
[...]

PR #165, opened 2017-01-11
Aeson 1.1

MichaelXavier:
This is for #162. Tests pass for me. We'll see what travis says. Note that I specifically force aeson 1.1 in stack-8.0.yaml to make sure its tested in aeson. It isn't in any actual stackage releases at the moment so you'll have to do something like this no matter what.

### DIFF
--- a/bloodhound.cabal
+++ b/bloodhound.cabal
@@ -32,7 +32,7 @@ library
   build-depends:       base             >= 4.3     && <5,
                        bytestring       >= 0.10.0  && <0.11,
                        containers       >= 0.5.0.0 && <0.6,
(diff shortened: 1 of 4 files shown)


bitemyapp:
@MichaelXavier I merged this because it looked fine, but should we have bumped the extra-dep in stack.yaml to test it in CI?

MichaelXavier:
The test matrix actually selects the appropriate stack yaml based on the GHC version, so the matrix entry for GHC 8 will exercise aeson-1.1 in travis. We could force it through to the other GHC versions but that will require a lot more version tetris in their stack.yaml files and I figure we're just after a smoke test anyways.

Issue #164, opened 2017-01-09
Sub-metrics parser


bermanjosh:
The `fromJSON` instances of the various "result" types for Aggregations (`TermsResult`, `DateHistogramResult`, `DateRangeResult`) all parse for sub-aggregations with a static key `"aggregations"` ( [here](https://github.com/bitemyapp/bloodhound/blob/244a123fd92ed7a20a77012ba84a5c81c1103612/src/Database/Bloodhound/Types.hs#L2034), [here](https://github.com/bitemyapp/bloodhound/blob/244a123fd92ed7a20a77012ba84a5c81c1103612/src/Database/Bloodhound/Types.hs#L2042) and [here](https://github.com/bitemyapp/bloodhound/blob/244a123fd92ed7a20a77012ba84a5c81c1103612/src/Database/Bloodhound/Types.hs#L2054)).

I think that might be a mistake. Nearest I can tell, sub-aggregations are returned with the name of the aggregation as the key. You can see an example from the [Elastic docs](https://www.elastic.co/guide/en/elasticsearch/guide/current/_adding_a_metric_to_the_mix.html), the returned JSON looks like:
<pre><code>  "aggregations": {
      "colors": {
         "buckets": [
            {
               "key": "red",
               "doc_count": 4,
               <b>"avg_price"</b>: { 
                  "value": 32500
               }
            }
         ]
      }
   }</code></pre>

I didn't see a test case for sub-aggregations (unless I missed it, which is totally possible). 

@chrisguiney @alistair @MichaelXavier @MHova : What do you think? Thanks

MHova:
Ah interesting @bermanjosh. I want to make sure that this is indeed a bug with the library. Can you write failing test cases that demonstrate this problem against a live ES instance for each of the major ES versions supported by `Bloodhound`?

bermanjosh:
@MHova I confirmed this locally while I was working on it today. What're you looking for, some branch pushed up for travis specifically to fail? Or just some minimal-code here? I'm happy to do either, I just wasn't clear which you meant. 

I've also started working on a rough solution, essentially just passing the object of any unused fields as a sub-aggregation, and letting the user call the aggregation name while making a Result. I'm building it off the state of my fork for #158, so I don't think I'll be able to PR it until after that's accepted / merged.

MHova:
[...]

Issue #163, opened 2017-01-08
Support for `top_hits` Metrics Aggregation


bermanjosh:
The 'Top hits Aggregation' is a way of ordering results within a bucket. I believe I need it for a new use case I'm working on (grouping similar search results). Based on what I can tell, it differs from other aggregations in 2 ways:
1. It returns the document as part of the aggregation bucket
2. It isn't "nestable", that is, it's only supposed to be used as a lower-level aggregation, not a top-level, and it cannot have sub-aggregations. 

I'd like to add support for this, specifically, hitting the following criteria:
- It's `AggregationResults` instance returns the actual document / Haskell object. It seems to be similar to `Hit a`, so I'm hoping I can piggyback on that. 
- Nice to have - Enforce the lower-level requirement. 

I might add more here as I work through it, if that's ok. 

Edit - @bitemyapp, can you assign this to me?

bermanjosh:
Tentative implamentation is in: https://github.com/bermanjosh/bloodhound/commits/badAggTest. Will PR after #158.

PR #168, opened 2017-01-30
Sub aggs

bermanjosh:
For #164 and #163 

- Makes the sub-aggregation parser more generic to allow for named aggregations. Adds a test for this case. 
- Support basic clauses of the `top_hits` aggregation.


bermanjosh:
Side question - is everyone ok with my changes to the stack.yaml? I'm on Windows, so symlinks are hit-and-miss. If not, feel free to switch it back, and I'll force my git to play nice.

bitemyapp:
I can't really tell from the Github diff what changed other than the nightly bump and extra dep removal.

I'll assume that you changed a symlink into a real file that is duplicating other content.

I am hypothetically fine doing that but you need to deduplicate the content.

That means no more stack-8.0.yaml, just a default `stack.yaml` and an alternate version. Makefile/.travis.yml will need to be corrected too.

I'll let you decide whether this is more or less hassle than fixing symlinks on your system.

PR otherwise looks okay.

bermanjosh:
> I'll assume that you changed a symlink into a real file that is duplicating other content.

Correct - I'll do the de-duplicating

bermanjosh:
Done - I wasn't sure what you're using the Makefile for, so check my changes. I set GHC 8.0 as the default and added 7.10.

PR #159, opened 2016-12-15
Fix "tie_breaker" to/fromJSON in MultiMatchQuery

bermanjosh:

PR #172, opened 2017-02-09
Forgot a tie_breaker in #159

bermanjosh:
I forgot an instance when I did #159

### DIFF
--- a/src/Database/V5/Bloodhound/Types.hs
+++ b/src/Database/V5/Bloodhound/Types.hs
@@ -2781,7 +2781,7 @@ instance ToJSON MultiMatchQuery where
                  , "query" .= query
                  , "operator" .= boolOp
                  , "zero_terms_query" .= ztQ
-                 , "tiebreaker" .= tb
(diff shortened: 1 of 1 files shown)


bitemyapp:
Fine with the change in principle, but:

1. It broke the build with GHC 8.0

2. I'd feel better if there was a test accompanying this, it's now the second time the same mistake has been made.

bermanjosh:
1) The GHC 8 build is broken, I grant you - but it's not because of this commit. It also happened for @MichaelXavier  in #171. It seems to be verbosely named type import. 
2) Sure

I've got a few work things I have to finish up over the next few days, but I'll try to get to both of these Sunday or Monday.

bitemyapp:
@bermanjosh @MichaelXavier I can repro this locally but the error makes little to no sense to me. I'd expect something like this if we were mixing package imports but we are not.

bermanjosh:
I'm going to be submitting a new PR shortly which supersedes this (and includes the missing test which was requested here).

PR #174, opened 2017-02-13
Drop doctests

MichaelXavier:
So I feel like I convered the meaningful text that was previously a doctest in the source code. As we discussed, I also removed all the code samples from the README which were never built and thus likely to lie. I didn't really port those into the examples file as a lot of them were not really sensible for a compiled example because they were just showing you how to use constructors. Still, I cut out some exposition as well in the process. I'm not sure what to do here. I could roll back changes to that file if anyone actually gets any use out of it but then we run the risk of the file not keeping up with the type changes.


MichaelXavier:
I should also mention that this incorporates the semigroups PR I made and brings us to tests passing on nightlies so bloodhound should be able to get back into stackage.

stack-7.10.yaml
```diff
@@ -1,13 +1,15 @@
-flags: {}
+flags:
+  bloodhound-examples:
+    werror: true
```
bitemyapp: veddy good

bitemyapp:
@MichaelXavier we need to drop the hspec upper bound as well, then it can get back into Stackage I think. I'll do that after merging this (when it passes)

bitemyapp:
Restarted build, sigh.

stack-7.10.yaml
MichaelXavier: I'm not devious enough to put it on the whole project yet but give it time :P

MichaelXavier:
@bitemyapp we're green!

PR #175, opened 2017-02-20
Sorted version of getInitialScroll for V5 client

WraithM:
Elasticsearch 2/5 seems to be able to do sorted scrolling searches. I wanted to make a sorted scroller, and this worked well for me.

### DIFF
--- a/src/Database/V5/Bloodhound/Client.hs
+++ b/src/Database/V5/Bloodhound/Client.hs
@@ -55,6 +55,7 @@ module Database.V5.Bloodhound.Client
        , searchByType
        , scanSearch
        , getInitialScroll
+       , getInitialSortedScroll
        , advanceScroll
        , refreshIndex
        , mkSearch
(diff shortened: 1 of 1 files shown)


MichaelXavier:
Are there any other variations of the scroll API we may support? This may be calling for a more configurable API.

WraithM:
Great question. At least for me, I haven't needed anything other than the sorting.

WraithM:
We could also merge my version of `getInitialScroll` with the original. We could put some flag on it to force sort by `_doc` or not.

WraithM:
Any thoughts on this? There don't seem to be any other unimplemented aspects of the scroll api. It just has that note about using _doc sort for optimizations.

https://www.elastic.co/guide/en/elasticsearch/reference/5.2/search-request-scroll.html

MichaelXavier:
@bitemyapp what do you think? Sounds like as it stands there are 2 variations of the scroll API that are as separate functions in the code. We could unify them and break API or keep the 2 functions, wait and see if more are added in future ES versions and then maybe do a unification and break then? I'm leaning towards the latter.

bitemyapp:
@MichaelXavier I prefer the latter meself.

PR #190, opened 2017-07-12
Add support for script-like things (`function_score` and `script_fields`)

lunaris:
This PR adds (`V5`) support for:

* `function_score` (https://www.elastic.co/guide/en/elasticsearch/reference/current/query-dsl-function-score-query.html), currently supporting inline scripts, stored scripts and all functions except decay functions.
* `script_fields` (https://www.elastic.co/guide/en/elasticsearch/reference/current/search-request-script-fields.html), returned in `hits` under `fields`.

Things missing/work-in-progress/issues:

* This is probably a breaking change since the `Script` type has been elaborated to support more than the `inline` case it previously exclusively catered for.
* Documentation for new types.
* `V1` support (not sure if required to get this merged).
* More comprehensive testing (currently I've added the bare-minimum round-trip JSON tests).

### DIFF
--- a/src/Database/V5/Bloodhound/Client.hs
+++ b/src/Database/V5/Bloodhound/Client.hs
@@ -1044,7 +1044,7 @@ scanSearch indexName mappingName search = do
 -- >>> mkSearch (Just query) Nothing
 -- Search {queryBody = Just (TermQuery (Term {termField = "user", termValue = "bitemyapp"}) Nothing), filterBody = Nothing, sortBody = Nothing, aggBody = Nothing, highlight = Nothing, trackSortScores = False, from = From 0, size = Size 10, searchType = SearchTypeQueryThenFetch, fields = Nothing, source = Nothing}
 mkSearch :: Maybe Query -> Maybe Filter -> Search
-mkSearch query filter = Search query filter Nothing Nothing Nothing False (From 0) (Size 10) SearchTypeQueryThenFetch Nothing Nothing Nothing
+mkSearch query filter = Search query filter Nothing Nothing Nothing False (From 0) (Size 10) SearchTypeQueryThenFetch Nothing Nothing Nothing Nothing
 
 -- | 'mkAggregateSearch' is a helper function that defaults everything in a 'Search' except for
 --   the 'Query' and the 'Aggregation'.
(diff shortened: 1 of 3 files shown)


bitemyapp:
>This is probably a breaking change since the Script type has been elaborated to support more than the inline case it previously exclusively catered for.

I don't mind major version bumps, my main practical concern is that no previously existing code should break at runtime. Is that the case? To that end, I think we need to start making all the datatypes strict so that people using record assignment syntax to construct these values don't get accidental bottoms.

@MichaelXavier What version of GHC are you on? Do you have any objections to flipping `StrictData` on or to moving things to strict fields?

>Documentation for new types.

This doesn't have to be elaborate, but a brief explanation with a link to the ES docs is always appreciated.

>V1 support (not sure if required to get this merged).

I mean, you're already running a fork, you don't actually need us to merge this for your needs to be served, so it seems to be like harmonizing the work with V1 is what would make a merge + release more worthwhile. I'll do it anyway if you twist my arm but I need to be consistent about enforcing this so that the APIs stay in sync. I'd prefer not to do a big catch-up cycle later to correct drift, we don't really have mechanical API diffing in Haskell. I'd feel differently if we did.

>More comprehensive testing (currently I've added the bare-minimum round-trip JSON tests).

ES have changed their minds in the past about how scripting and the script language should work. A functional test is the only real assurance we have so this is going to be necessary for me to merge the PR.

Please let me know if you have any questions or need assistance.

MichaelXavier:
@bitemyapp I think the oldest project we have is on 7.10.3 but it may be up to 8 even. Everything new I use the latest LTS at the time and upgrade it frequently. But maybe I'm not the best to ask. I also don't care about strict fields but I'm not sure what that has to do with this PR. I can't figure out from GHC Trac when StrictData arrived, but it seems like either in 7.8 or 7.11 which I didn't know was a thing.

bitemyapp:
@MichaelXavier >I also don't care about strict fields but I'm not sure what that has to do with this PR.

Apologies, I should've made a separate issue. I thought of accidental bottoms in records as being an example of how adding fields could cause things to break at runtime. An unassigned strict field in a record created using the record update syntax is a compile-time error.

MichaelXavier:
Ah good point. I use -Wall -Werror on everything (and I'm taking a crack at the ticket that does that for this repo) so I usually get the error either way.

bitemyapp:
V5-only, but I've manually integrated this PR with the faster compilation work I was doing in this branch: https://github.com/bitemyapp/bloodhound/tree/faster-compile-script-merge

With that, I consider this PR pseudo-merged so I'll close it now.

Issue #194, opened 2017-07-12
Consider workaround for stalled builds


MichaelXavier:
I just dealt with this in a project that depends on bloodhound. I think because the bloodhound package itself take so long to compile, combined with the probably pretty slow hardware on travis free, we often stall out because there's no output after 10 minutes. I added the following to my project's `install` stanza in `.travis.yaml`: `- while true; do echo "travis pls"; sleep 30; done &` and it worked. I figured I'd ask here before doing something so ridiculous on this project.

bitemyapp:
I think this is fixed now. All hail `stack build --fast` and 10gb of TravisCI cache.

PR #195, opened 2017-07-13
Add travis output to prevent stalling

MichaelXavier:
This is for #194

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -31,6 +31,8 @@ install:
   # set up a repo for snapshot testing. Required in ES >= 1.6
   - echo "path.repo = [\"/tmp\"]" >> ./elasticsearch-$ESVER/elasticsearch.yml
   - ./elasticsearch-$ESVER/bin/elasticsearch &
+  # Ensure constant output so we don't sometimes get killed for taking too long to compile without output
+  - while true; do echo "travis pls"; sleep 30; done &
 
 script:
   - stack setup
--- a/.travis.yml
+++ b/.travis.yml
@@ -21,7 +21,7 @@ env:
 install:
   # stack
   - mkdir -p ~/.local/bin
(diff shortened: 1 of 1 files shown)


bitemyapp:
@MichaelXavier I think `travis_wait` works for the 10 minute output case? Does the busy-spinner do something TravisCI's provided thingy would not?

https://docs.travis-ci.com/user/common-build-problems/#Build-times-out-because-no-output-was-received

This PR build failed because we hit the 50 minute global limit on a couple of the jobs.

[You won't like the answer](https://github.com/travis-ci/travis-ci/issues/2736#issuecomment-109778499)

bitemyapp:
For an idea of the speed-gap between dedis and AWS-based CI services like CircleCI and TravisCI: http://bitemyapp.com/posts/2016-03-28-speeding-up-builds.html

bitemyapp:
I am giving this a crack right now.

bitemyapp:
https://github.com/travis-ci/travis-ci/issues/7020 christ on sale

bitemyapp:
screen shot 2017-07-13 at 1 36 31 am
screen shot 2017-07-13 at 2 06 27 am

should be g2g now, merge if you're happy. imma pass tf out now

MichaelXavier:
LOL. I wasn't aware of travis_wait. That's just like me for a system to report an error with a page to read more and for me to not visit that page and come up with some much dumber solution. This seems like a better solution, thanks!

bitemyapp:
@MichaelXavier FWIW, there's a trade-off here. I halved our build time but our cache payloads are _huge_ (~6gb per PR/branch) now, but without having done so, our builds wouldn't have been passing anyway. That means cache upload period at end of build is longer. I think it's worth it, but keep it in mind in case we have problems later.

The other thing I did to speed up our builds was turn on `--fast` for build and test.

MichaelXavier:
Good point. Yeah since we're not building any artifacts with our CI, may as well use --fast to speed things up slightly.

PR #198, opened 2017-08-06
Fix improve example

yannick-cw:
Hi, I had a hard time figuring out how to parse the responses for searches.
When trying to run the example I also found out some stuff was not working :)

I'd like to add an example here for parsing and I fixed the geo object datatype, as it only accepts `lat`, `lon` now as field names.

I am just starting to learn haskell, so please forgive me if something is bad style.

### DIFF
--- a/examples/Tweet.hs
+++ b/examples/Tweet.hs
@@ -32,19 +32,12 @@ instance ToJSON TweetMapping where
 
 
 -------------------------------------------------------------------------------
-data Location = Location
-  { locLat :: Double
-  , locLon :: Double
-  } deriving (Eq, Generic, Show)
-
-
--------------------------------------------------------------------------------
 data Tweet = Tweet
   { user     :: Text
   , postDate :: UTCTime
   , message  :: Text
   , age      :: Int
-  , location :: Location
+  , location :: LatLon
   } deriving (Eq, Generic, Show)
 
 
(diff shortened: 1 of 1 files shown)


examples/Tweet.hs
```diff
@@ -108,7 +97,10 @@ main = runBH' $ do
   let boost = Nothing
```
yannick-cw: The `putTemplate` is failing as well at the moment with `Root mapping definition has unsupported parameters:  [location : {type=geo_point}]\"}]`

I could not figure out how to fix it, my only working fix was to just leave the list of `TweetMapping` empty..

jagare: Hallo Yannick,
I run into the same problems yesterday.
Missed your PR before I did my own identical [PR #204].
I solved the putTemplate with [PR #205 ]

Cheers
/Lars

yannick-cw: Ah nice mapping fix makes sense, I only wonder why the mapping worked with the `   _ <- putMapping testIndex testMapping TweetMapping
`

jagare: Well, as it was before the change I think it is according to ES spec for  [Put Mapping case 2](https://www.elastic.co/guide/en/elasticsearch/reference/current/indices-put-mapping.html).
 That is, the mapping type is given by the URL and need not be specified in the JSON payload. However with my change of the mapping to work with the [PR #205 ] it is not really according to the manual, since the type name is defined twice. It works, but maybe not perfectly correct.

bitemyapp:
Thank you for putting this up! I apologize for not getting to it sooner :cry: 

I merged the other PR which fixed this, so hopefully nobody else trips over this.

yannick-cw:
@bitemyapp perfect :)
What do you think of extending the example file with the reply parsing?
```haskell
 reply <- searchByType testIndex testMapping search
 Right searchResult <- parseEsResponse reply :: BH IO (Either EsError (SearchResult Tweet))
 let tweets = fmap hitSource (hits $ searchHits searchResult) >>= maybeToList
 liftIO (print tweets)
```

bitemyapp:
@yannick-cw I think it'd be a good idea for it to do that and to detect errors, throwing an exception if one occurs.

Issue #148, opened 2016-11-02
aeson Encoding for faster Bulk operations


andrewthad:
Currently, `Bulk` is defined as:

    data BulkOperation =
        BulkIndex  IndexName MappingName DocId Value
      | BulkCreate IndexName MappingName DocId Value
      | BulkDelete IndexName MappingName DocId
      | BulkUpdate IndexName MappingName DocId Value

I would like to add three new data constructors:

    data BulkOperation =
        BulkIndex  IndexName MappingName DocId Value
      | BulkCreate IndexName MappingName DocId Value
      | BulkDelete IndexName MappingName DocId
      | BulkUpdate IndexName MappingName DocId Value
      | BulkIndexEncoding  IndexName MappingName DocId Encoding
      | BulkCreateEncoding IndexName MappingName DocId Encoding
      | BulkUpdateEncoding IndexName MappingName DocId Encoding

These would use the `Encoding` type from `aeson-0.11` and onward. This would allow me to perform faster bulk operations `ToJSON` types that implement `toEncoding`. Would a PR with this change (and appropriate changes to the implementation of `bulk`) be accepted?

bitemyapp:
Also potentially relevant is https://github.com/bitemyapp/bloodhound/pull/145

Your approach seems like it avoids breaking things for most users. Have you benchmarked that it makes things faster? Can you share anything from that?

/cc @MichaelXavier

MichaelXavier:
I'm interested in the benchmarks too. I do like that this is backwards compatible, but I'm kind of curious how much speed there is to gain going from Value to Encoding and especially how relevant it is in the face of making network calls to ES (obviously you can mitigate this quite a bit by batching). Also, if we did go forward with this, it would be a good idea to document on the BulkOperation type why you'd use one over the other.

andrewthad:
I haven't done any benchmarking yet. I had seen the note in the docs that bulk had been changed from taking a List to taking a Vector to reduce allocations, and I assumed that this would take it a step further. I don't have the time to put together a meaningful benchmark any time soon, but I can definitely cut a PR if you'd take it without evidence of better performance. Also, I'm probably going to try this out in a proprietary codebase in three weeks, so I can let you know the results of that, but I think factoring out a standalone benchmark that other people can run would be to time consuming.

MichaelXavier:
[...]

PR #201, opened 2017-08-07
bulk create encoding

andrewthad:
resolves https://github.com/bitemyapp/bloodhound/issues/148


bitemyapp:
Could you add a test for this please?

andrewthad:
@bitemyapp The test has arrived.

PR #204, opened 2017-08-10
Use Types.LatLon for location

jagare:
Without this change, code compile but fail against ES, since names for lat lon (in Location record) don't follow specification for geo_point. I changed to use Types.LatLon instead.

There is still a problem with the example regarding the IndexTemplate. However that is not addressed in this PR.


jagare:
Realized the same problem is present in the tests. I'll have a look at that to (for ES5).

jagare:
Now I found [PR #198] with an identical solution as far as I can see. Guess this one can be abandoned.

PR #205, opened 2017-08-10
Examples TweetMapping fix structure

jagare:
The Tweet example failed due to TweetMapping not having the proper structure.
The error in ES5 is: 

> failed to put template [tweet-tpl]
org.elasticsearch.index.mapper.MapperParsingException: Failed to parse mapping [properties]: Root mapping definition has unsupported parameters:  [location : {type=geo_point}]

### DIFF
--- a/examples/Tweet.hs
+++ b/examples/Tweet.hs
@@ -25,9 +25,11 @@ data TweetMapping = TweetMapping deriving (Eq, Show)
 
 instance ToJSON TweetMapping where
   toJSON TweetMapping =
-    object
-      [ "properties" .=
-        object ["location" .= object ["type" .= ("geo_point" :: Text)]]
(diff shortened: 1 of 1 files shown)


jagare:
Hmmm, maybe this is not correct since TweetMapping is also used for putMapping.
According to ES spec [Put Mapping](https://www.elastic.co/guide/en/elasticsearch/reference/current/indices-put-mapping.html#indices-put-mapping) can be used in several ways. The current implementation correspond to case 2 and 3 in the ES docs. Then the mapping type, ie "tweet" in the example, should not be included, since it is part of the URL.
Guess I should modify this.

bitemyapp:
Closing for now as we are reorganizing the library internally, let me know if you'd like to revisit this.

PR #206, opened 2017-08-10
ES5 default for tests

jagare:
It make more sense to have ES5 as default for tests rather than ES1.


bitemyapp:
Agreed, thank you!

PR #216, opened 2017-12-05
add ability to change compression codec. This allows us to use DEFLATE instead of lz4.

andrewthad:
…E instead of lz4.

### DIFF
--- a/src/Database/V5/Bloodhound/Types.hs
+++ b/src/Database/V5/Bloodhound/Types.hs
@@ -74,6 +74,7 @@ module Database.V5.Bloodhound.Types
        , UpdatableIndexSetting(..)
        , IndexSettingsSummary(..)
        , AllocationPolicy(..)
+       , Compression(..)
        , ReplicaBounds(..)
        , Bytes(..)
        , gigabytes
@@ -586,6 +587,7 @@ data UpdatableIndexSetting = NumberOfReplicas ReplicaCount
                            | TTLDisablePurge Bool
                            -- ^ Disables temporarily the purge of expired docs.
                            | TranslogFSType FSType
+                           | CompressionSetting Compression
                            | IndexCompoundFormat CompoundFormat
                            | IndexCompoundOnFlush Bool
                            | WarmerEnabled Bool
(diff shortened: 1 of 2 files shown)


bitemyapp:
Are you trying to save CPU time? I'm curious what triggered this.

andrewthad:
Other way around. I'm trying to save disk save. According to the [disk tuning docs](https://www.elastic.co/guide/en/elasticsearch/reference/current/tune-for-disk-usage.html#_use_literal_best_compression_literal):

> The _source and stored fields can easily take a non negligible amount of disk space. They can be compressed more aggressively by using the best_compression codec.

src/Database/V5/Bloodhound/Types.hs
```diff
@@ -600,6 +602,20 @@ data ReplicaBounds = ReplicasBounded Int Int
                    | ReplicasUnbounded
                    deriving (Eq, Read, Show, Generic, Typeable)
 
+data Compression = CompressionDefault | CompressionBest
```
bitemyapp: I realize these data constructors are representative of the underlying text in the JSON but I'd like some documentation of what compression algorithm they are likely to force in the Lucene index and why you'd want one versus the other.

andrewthad:
I've improved the docs. Also, I forgot to mention that the feature added in this PR cannot be used without first merging https://github.com/bitemyapp/bloodhound/pull/209. This index setting isn't cannot actually be updated dynamically, so it needs `createIndexWith` from the other PR.

bitemyapp:
@andrewthad could you add the compression setting to the `createIndexWith` test case? If Elastic changes the API I'd like the tests to go red.

andrewthad:
when has elastic ever made haphazardly documented breaking changes to their api? ;)
I'll close this PR and rebase on top of https://github.com/bitemyapp/bloodhound/pull/209 and then add the tests there.

andrewthad:
Ah, you just merged it. Nevermind, I'll just rebase this on top of master and add the test.

bitemyapp:
Gah, sorry if I'm making this difficult for you!

andrewthad:
No worries. There's now a test for both compression codecs.

bitemyapp:
@andrewthad okay, should I merge this one then?

andrewthad:
Yes, this one's got everything now.

bitemyapp:
@andrewthad thank you for everything!

PR #218, opened 2017-12-13
Direct Generators support

bermanjosh:
This PR adds the ability to modify the `Direct Generators` object within the `Phrase Suggester`. 

[Docs link](https://www.elastic.co/guide/en/elasticsearch/reference/5.6/search-suggesters-phrase.html#_direct_generators)


bermanjosh:
@MichaelXavier @bitemyapp - just bumping this - thanks.

src/Database/V5/Bloodhound/Types.hs
```diff
@@ -5293,6 +5296,7 @@ data PhraseSuggester =
                   , phraseSuggesterShardSize :: Maybe Int
                   , phraseSuggesterHighlight :: Maybe PhraseSuggesterHighlighter
                   , phraseSuggesterCollate :: Maybe PhraseSuggesterCollate
+                  , phraseSuggesterCandidateGenerators :: [DirectGenerators]
```
MichaelXavier: Wouldn't the addition of this field to an exported type constitute a B version number change instead of a C?

MichaelXavier:
Had one minor comment about it possibly being a breaking change since it technically adds a field to a fully exported constructor. Also curious if this should be added to the V1 and V2 modules tree as well, a cursory look seems like [direct generators were around in 1.x](https://www.elastic.co/guide/en/elasticsearch/reference/1.7/search-suggesters-phrase.html#_direct_generators).

Beyond that, this falls into my normal evaluation of an ES feature addition: I don't really know what this does but as long as it works and doesn't block future feature additions, seems good to me.

bermanjosh:
@MichaelXavier I ported the changes to V1 and bumped the version to 0.16. 

All the old tests are passing, but 8.2 isn't. That's a separate problem, I don't think this PR is the right place to deal with it. Let me know if you disagree.

bitemyapp:
@bermanjosh @MichaelXavier  wait on the merge until I can unbreak master. Should've put that in a topic branch. Thought a GHC upgrade wouldn't be a big deal. This is why we don't make exceptions to the rules eh?

bitemyapp:
Pull from master, it's clean now.

bermanjosh:
@bitemyapp Done & working - thanks

bermanjosh:
@bitemyapp Bump - thanks.

bitemyapp:
@bermanjosh Thank you!

bermanjosh:
Thanks to you!

PR #221, opened 2018-01-29
Allow http-types-0.12

MichaelXavier:
I tested this locally, but I'm not really sure its the right thing to
put http-types in the default stack.yaml

This is for https://github.com/fpco/stackage/issues/3232 which blocks
us in katip-elasticsearch downstream

Update: @bitemyapp I noticed you actually put the http-types-0.11 in the stack.yaml in the last commit. Should I instead put 0.12 in the stack.yaml? Wasn't sure about the policy but I may have guessed wrong.

### DIFF
--- a/bloodhound.cabal
+++ b/bloodhound.cabal
@@ -55,7 +55,7 @@ library
                        text             >= 0.11    && <1.3,
                        mtl              >= 1.0     && <2.3,
                        transformers     >= 0.2     && <0.6,
-                       http-types       >= 0.8     && <0.12,
+                       http-types       >= 0.8     && <0.13,
                        vector           >= 0.10.9  && <0.13,
                        scientific       >= 0.3.0.0 && <0.4.0.0,
                        exceptions,
--- a/stack-8.2.yaml
+++ b/stack-8.2.yaml
@@ -6,7 +6,6 @@ packages:
   - '.'
   - './examples'
 
-extra-deps:
-  - http-types-0.11
+extra-deps: []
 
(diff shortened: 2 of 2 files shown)


bitemyapp:
>I noticed you actually put the http-types-0.11 in the stack.yaml in the last commit. Should I instead put 0.12 in the stack.yaml? Wasn't sure about the policy but I may have guessed wrong.

Yeah, to test that we're correctly tracking Stackage and that the build works I put http-types-0.11 in the `stack.yaml`. Part of the intent is to pin the latest version we're intending to work with in whatever the primary stack.yaml is at the moment. This build permits http-types-0.12 it isn't validating that the build works with that specific version. I'm quite open to alternatives.

MichaelXavier:
That sounds like a good tradeoff to me. I'll add an explicit http-types version and see what the build says.

MichaelXavier:
@bitemyapp Tests are green. Mind if I cut, I guess, a D-level release?

bitemyapp:
@MichaelXavier much as I'm loathe to break things up, I think we're eating non-linear `Generics` pain in the Types modules. Might it be worth to split them into constituent modules to ease compile times? I hate bouncing the CI on this library.

bitemyapp:
>Mind if I cut, I guess, a D-level release?

Yeah go ahead, thanks for covering this!

MichaelXavier:
I've created #222 to discuss the compile time issue.

MichaelXavier:
Released at 0.15.0.2

PR #223, opened 2018-01-31
[WIP] Getting faster compiles out of Bloodhound

bitemyapp:
Speaking casually from using `ghcid` (which I strongly recommend when ripping things apart like this) it seems faster now. YMMV.

Lots more work to do:

- The library and examples build, the tests do not.

- Finish purging `Generic` by addressing `ApproxEq` and whatever else wanted it

- Replace QuickCheck with Hedgehog (should probably do this one in a separate branch based on this one for safety's sake?)


bitemyapp:
More broadly, how do you feel about V1 support? Do you know anyone that still uses V1?

bitemyapp:
```
$ time make
stack build
bloodhound-0.15.0.2: unregistering (local file changes: bloodhound.cabal)
bloodhound-0.15.0.2: configure (lib)
bloodhound-0.15.0.2: build (lib)
bloodhound-0.15.0.2: copy/register
bloodhound-examples-0.0.0.0: configure (exe)
bloodhound-examples-0.0.0.0: build (exe)
bloodhound-examples-0.0.0.0: copy/register
Completed 2 action(s).
Log files have been written to: /Users/callen/work/bloodhound/.stack-work/logs/
make  8.23s user 3.15s system 76% cpu 14.902 total
```

d4a4764

MichaelXavier:
I'm hoping I'll have some time to review this tonight. As for V1, I'm still using it, unfortunately. Provisioning a new cluster just never ends up in the sprints. However, I'm willing to step up work on bringing V1 up to date when people existing in the present day do work on the V5 module. You could also maybe soft deprecate V1 and mention in the README or haddocks that it may be behind. If folks come across a feature in V5 they need in V1, we could merely accept PRs for it. But then we could stop asking others adding features to V5 to do extra labor on a version they haven't touched in years.

andrewthad:
I've added several features to V5, and I haven't been asked to add them to V1, so I think that bridge might already be crossed :)

Also, just to get another idea out there about V1, it could be moved into a separate package.

bitemyapp:
@MichaelXavier that's appreciated.

Here's my proposal/request:

- Prototype this refactoring with the V5 variant (I'm willing to continue hammering at this)
- When we've arrived at a final version that we're happy with, replicate the work to ameliorate compile times with the V1 variant. This is where I'd ask you to be willing to take that on. I realize we mostly touch V5 on a regular basis but I care about first-time user experience WRT compile times as well.

Is that something you're okay doing? If not I'd be happy to know what works for you.

@andrewthad I've asked other contributors in the past and been ignored, often not returning to the PR so I stopped bothering. Given how challenging it can be to get contributions into a mergeable state from people we'd (Michael and I) need to eat the replication work ourselves. To their credit, some contributors did include or add parallel V1/V5 changes in their patches and I was grateful for it.

MichaelXavier:
Yes, I'd be willing to replicate the work to V1.

Also, I'd be fine with cutting V1 loose into its own package if it would be worthwhile for reducing work for contributors and for the library itself. To be completely upfront: I would probably leave the V1 package alone most of the time but would merge PRs against it if other people had more pressing needs. I'm hoping sometime soon I will be able to get off V1 and after that point I'd probably only contribute custodial work (merging PRs and cutting releases) to V1.

bitemyapp:
@MichaelXavier What you'll likely want to know about the most recent batch of changes to the tests is that I neutered `ApproxEq` to be scoped only to the types you actually wanted to validate that for and removed all Generics.

MichaelXavier:
I'm all for it. ApproxEq definitely was due to be reconceived as a solution.

andrewthad:
It would be helpful to contributors to merge existing PRs before redoing the test suite. Otherwise, I'll need to rewrite passing tests in my open PRs.

bitemyapp:
@andrewthad I think it'd be faster for me to merge and fix peoples' PRs myself assuming there aren't domain specific problems.

andrewthad:
I was referring to the older PRs, not ones I made today. But if you don't mind fixing them yourself, then by all means, do that.

MichaelXavier:
@bitemyapp I won't guess how long it will take but I think I should be able to give it a try on Saturday. Do I just start from `we-need-faster-compile`?

bitemyapp:
>Do I just start from we-need-faster-compile?

Yep! I'm currently modularizing and cleaning up V5's tests.

MichaelXavier:
I sat down to take a look at this and had a few questions. I was planning on approaching this pretty mechanically: pretty much copying your structure from V5, sorting out any issues that come up (I'm sure V5 has types that aren't in V1), then do the same with tests so we have basically mirrored modules again.

1. I noticed the `V5.Types.Internal` module isn't exported anymore but `V1` was. Is this on purpose?
2. I noticed the main `V5.Types` module has a lot of orphaned instances, presumably from the types being moved to subdir modules and the instances not following. Is this a compilation speed optimization or should these instances be moved to their respective subdir modules?
3. Generally speaking aside from these issues (which I'm happy to even resolve today on this branch, since I've got a bit of time), is the `src` tree basically done? I'd like to avoid a bunch of resynchronization of changes when V1 and V5 come together.

bitemyapp:
>I noticed the V5.Types.Internal module isn't exported anymore but V1 was. Is this on purpose?

`V5/Internal.hs` doesn't exist as an independent module anymore. I suppose I could reconstitute it? It was always marked internal and I'm not sure what the benefit of recreating it would be. I think I'd rather add the Internal sub-modules to the exports so people can get at stuff if they need to. What do you think?

>I noticed the main V5.Types module has a lot of orphaned instances

Yeah this is my bad, I need to do a `-Werror` pass.

>Generally speaking aside from these issues (which I'm happy to even resolve today on this branch, since I've got a bit of time), is the src tree basically done? I'd like to avoid a bunch of resynchronization of changes when V1 and V5 come together.

Not yet, I'll let you know. Sorry for the delay!

MichaelXavier:
> V5/Internal.hs doesn't exist as an independent module anymore. I suppose I could reconstitute it? It was always marked internal and I'm not sure what the benefit of recreating it would be. I think I'd rather add the Internal sub-modules to the exports so people can get at stuff if they need to. What do you think?

I don't think we're obligated to maintain the presence of `Internal.hs` (though I'm unclear if the spirit of PVP applies to internal modules, i.e. do you need a breaking version change for removing it or modifying it significantly). I'd only say that if there are internals that we do want to reserve the right to modify at will or are caveat emptor, moving them to modules without `Internal` may mean we are obligated to think of them with respect to PVP (though of course, see question above).

bitemyapp:
@MichaelXavier test suite breakage is tied to harmonizing and fixing the test suite for the change-over.

I've reviewed the state of the V5 library and the test suite and I think I'm happy with this. I'll take responsibility for making any future duplicate changes if I notice anything. I also tried to move instances to live with the datatypes they were for in the modules but I may have missed a couple. Kicking the ball over to you now, let me know if you need anything.

MichaelXavier:
@bitemyapp I think I've got everything ready on `we-need-faster-compile-v1`. I basically took your module and test structure, pruned back bits that weren't in V1. This required a lot of piecemeal copy/pasting of code so the V1 branch looks like a truck drove through it but it compiles and seems to only fail a few configurations in the test matrix due to the same dependency reasons as your branch. LMK what you want me to do with the code.

bitemyapp:
@MichaelXavier is it up in a branch? I'd probably start a 5+1 branch, pull in this one, pull in yours, then start hammering things out for merge into this one and a final merge.

MichaelXavier:
@bitemyapp yes, the `we-need-faster-compile-v1` branch on this repository.

PR #236, opened 2018-05-03
Add wildcard query support

ahodgen:


bitemyapp:
Thank you for the contribution! Could you please add tests and add your name and what you did to the changelog document?

bitemyapp:
Thanks!

Issue #237, opened 2018-05-05
getNodesInfo failing on Elasticsearch 5.6.3


bitemyapp:
Fails to parse the json below.

```json
{
  "_nodes": {
    "total": 1,
    "successful": 1,
    "failed": 0
  },
  "cluster_name": "elasticsearch",
  "nodes": {
    "jei7P8l5R46MZ5uiyNQ_zQ": {
      "name": "jei7P8l",
      "transport_address": "127.0.0.1:9300",
      "host": "127.0.0.1",
      "ip": "127.0.0.1",
      "version": "5.6.3",
      "build_hash": "1a2f265",
      "total_indexing_buffer": 207775334,
      "roles": [
        "master",
        "data",
        "ingest"
      ],
      "settings": {
        "client": {
          "type": "node"
        },
        "cluster": {
          "name": "elasticsearch"
        },
        "http": {
          "type": {
            "default": "netty4"
          }
        },
        "node": {
          "name": "jei7P8l"
        },
        "path": {
          "logs": "/Users/callen/work/elasticsearch/elasticsearch-5.6.3/logs",
          "home": "/Users/callen/work/elasticsearch/elasticsearch-5.6.3"
        },
        "transport": {
          "type": {
            "default": "netty4"
          }
        }
      },
      "os": {
        "refresh_interval_in_millis": 1000,
        "name": "Mac OS X",
        "arch": "x86_64",
        "version": "10.13.4",
        "available_processors": 8,
        "allocated_processors": 8
      },
      "process": {
        "refresh_interval_in_millis": 1000,
        "id": 79963,
        "mlockall": false
      },
      "jvm": {
        "pid": 79963,
        "version": "9.0.4",
        "vm_name": "Java HotSpot(TM) 64-Bit Server VM",
        "vm_version": "9.0.4+11",
        "vm_vendor": "Oracle Corporation",
        "start_time_in_millis": 1525548096016,
        "mem": {
          "heap_init_in_bytes": 2147483648,
          "heap_max_in_bytes": 2077753344,
          "non_heap_init_in_bytes": 7667712,
          "non_heap_max_in_bytes": 0,
          "direct_max_in_bytes": 0
        },
        "gc_collectors": [
          "ParNew",
          "ConcurrentMarkSweep"
        ],
        "memory_pools": [
          "CodeHeap 'non-nmethods'",
          "Metaspace",
          "CodeHeap 'profiled nmethods'",
          "Compressed Class Space",
[...]

PR #238, opened 2018-05-05
Semverification

bitemyapp:
Data.Version was breaking on my JVM version because it had a metadata build segment `+11`.

This fixes https://github.com/bitemyapp/bloodhound/issues/237

I decided to SemVer everything rather than just the JVM version parsing. `Data.Version` is pretty grody anyway. This handles V5 only.


bitemyapp:
`"vm_version\":\"25.171-b11\"`

tarnation

PR #242, opened 2018-07-31
Add support for Search After API

shulhi:
This adds support for https://www.elastic.co/guide/en/elasticsearch/reference/current/search-request-search-after.html

I decided to make the type `search_after` as `[Aeson.Value]` because this can be anything based on what user specified in `sort` body (in `Search`). `SearchAfterKey` is just type synonym to `[Aeson.Value]`.

Please let me know if this is okay. We are actually using this in production soon at [Plow](https://github.com/plow-technologies/) and it would be great if this PR can be merged so we can use the upstream version instead of our forked version.

### DIFF
--- a/src/Database/V5/Bloodhound/Client.hs
+++ b/src/Database/V5/Bloodhound/Client.hs
@@ -523,13 +523,13 @@ createIndex indexSettings (IndexName indexName) =
 createIndexWith :: MonadBH m
   => [UpdatableIndexSetting]
   -> Int -- ^ shard count
-  -> IndexName 
+  -> IndexName
   -> m Reply
 createIndexWith updates shards (IndexName indexName) =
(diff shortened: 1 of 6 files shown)


shulhi:
@bitemyapp it looks like the tests are failing for v1 with GHC8.0. It was roundtrips error for `BoolQuery` and I am not sure why. That part of code doesn't change and my changes should only affect v5. Search after API is not supported in V1. I might need some help with this.

MichaelXavier:
I kicked the test to run again. These roundtrip tests occasionally fail and I think they probably do with some better shrinking since in this case the repro query was massive. Nevertheless, if it is just a V1 failure and your feature isn't even available on V1, I'd say the failure shouldn't impede merging.

bitemyapp:
@shulhi Thank you for PR'ing this!

PR #245, opened 2018-10-27
Fix error in example code

ptillemans:
The example code dies with an error 

```
Response {responseStatus = Status {statusCode = 200, statusMessage = "OK"}, responseVersion = HTTP/1.1, responseHeaders = [("content-type","application/json; charset=UTF-8"),("content-encoding","gzip"),("content-length","131")], responseBody = "{\"_index\":\"twitter\",\"_type\":\"tweet\",\"_id\":\"1\",\"_version\":2,\"result\":\"updated\",\"_shards\":{\"total\":1,\"successful\":1,\"failed\":0},\"created\":false}", responseCookieJar = CJ {expose = []}, responseClose' = ResponseClose}
tweet-example: user error (Pattern match failure in do expression at Tweet.hs:89:3-6)
```

this is caused when verifying if the template exists after creating it.

The template creation fails because ES does not find the location parameter in the Root mapping.

```
root_cause": [
            {
                "type": "mapper_parsing_exception",
                "reason": "Root mapping definition has unsupported parameters:  [location : {type=geo_point}]"
            }
        ],
```

adding the name to the mapping fixes this and the example runs without errors.

### DIFF
--- a/examples/Tweet.hs
+++ b/examples/Tweet.hs
@@ -83,7 +83,7 @@ main = runBH' $ do
   True <- indexExists aliasName
 
   -- create a template so that if we just write into an index named tweet-2017-01-02, for instance, the index will be automatically created with the given mapping. This is a great idea for any ongoing indices because it makes them much easier to manage and rotate.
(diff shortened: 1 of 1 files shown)


bitemyapp:
Thank you!

PR #246, opened 2018-11-20
Relax the dependency on containers and allow building the test suite with GHC 8.6

23Skidoo:
NB: This PR drops support for GHC < 8. I'm not sure whether it is okay or not. Looks like there is no CI for GHC < 8, though.

Would be also nice if the dependency on `containers` was relaxed on Hackage or a new minor release with support for GHC 8.6 was uploaded.

### DIFF
--- a/bloodhound.cabal
+++ b/bloodhound.cabal
@@ -63,7 +63,7 @@ library
   other-modules:       Bloodhound.Import
                        Database.Bloodhound.Common.Script
   hs-source-dirs:      src
-  build-depends:       base             >= 4.3     && <5,
+  build-depends:       base             >= 4.9     && <5,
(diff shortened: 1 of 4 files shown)


MichaelXavier:
@bitemyapp i kicked the build and it passed. Any objection to dropping GHC 7.x support? If not, I can prepare a release. @23Skidoo this would constitute a C version bump as in A.B.C, right? Types didn't change but a newer version of dependencies is now required, which may introduce orphan instances or whatever.

23Skidoo:
Yep, there are no orphan instances in `containers` AFAICT.

23Skidoo:
Ping.

bitemyapp:
Fixed by #258. No upper bounds now.

PR #250, opened 2019-02-27
Add 'BulkUpsert' operation that allows for extra metadata.

Kheldar:
Hey - my team and I found that for our use-case, it was necessary to attach more metadata to `Update` actions than what seemed possible with the existing interface, so we added a new one with the fields we needed.

I didn't want to attach this stuff to the existing `BulkUpdate` operation, since I'm fairly sure that would be a breaking change.

If there's a better way to structure this or do this, let me know. We've been maintaining this branch for quite a while, so things in the library may have shifted since it was first written. :)

### DIFF
--- a/src/Database/Bloodhound/Client.hs
+++ b/src/Database/Bloodhound/Client.hs
@@ -911,6 +911,14 @@ mkBulkStreamValueAuto operation indexName mappingName =
           object [ "_index" .= indexName
                  , "_type"  .= mappingName]]
 
+mkBulkStreamValueWithMeta :: [UpsertActionMetadata] -> Text -> Text -> Text -> Text -> Value
+mkBulkStreamValueWithMeta meta operation indexName mappingName docId =
+  object [ operation .=
+          object ([ "_index" .= indexName
+                  , "_type"  .= mappingName
+                  , "_id"    .= docId]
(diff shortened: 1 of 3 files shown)


AlexeyRaga:
The build failure doesn't seem to do anything with @Kheldar's changes

AlexeyRaga:
@bitemyapp any chance you could look at this PR? It would be nice to have it in the official release instead of using a custom fork...

Kheldar:
The force pushes were to force rebuilds - a couple of the existing integration tests seem a little flakey.

AlexeyRaga:
@bitemyapp is there anything we can do to make this PR accepted? Any remarks about it?

bitemyapp:
@AlexeyRaga @Kheldar Apologies for the delay, I'll re-review soon.

bitemyapp:
I apologize for going AWOL and being so late to reply.

Having looked at it, the blocking issue is the lack of test cases.

Keeping this stuff working against Elasticsearch is a waking nightmare and I'm even more paranoid about functionality I didn't add myself. Elastic has a track record of churning API surface area and they don't publish a specification that would permit me to generate bindings the way Brendan Hay does for the amazonka family of libraries.

PR #264, opened 2019-10-15
Add 'BulkUpsert' operation (now with tests)

Kheldar:
Supercedes #250 

This adds an extra `Bulk` operation, with upsert semantics and the ability to pass some extra metadata that isn't possible with the existing API.

We've also added some tests for it, so hopefully it should be maintainable :)

### DIFF
--- a/src/Database/Bloodhound/Client.hs
+++ b/src/Database/Bloodhound/Client.hs
@@ -52,6 +52,7 @@ module Database.Bloodhound.Client
        , getDocument
        , documentExists
        , deleteDocument
+       , deleteByQuery
        -- ** Searching
        , searchAll
        , searchByIndex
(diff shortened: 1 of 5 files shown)


src/Database/Bloodhound/Client.hs
```diff
@@ -495,9 +497,9 @@ getNodesStats sel = parseEsResponse =<< get =<< url
   where
     url = joinPath ["_nodes", selectionSeg, "stats"]
     selectionSeg = case sel of
-      LocalNode -> "_local"
+      LocalNode          -> "_local"
```
bitemyapp: In future please leave whitespace/formatting changes out of a PR that makes semantic changes to the library.

bitemyapp: Not necessarily opposed to such changes, but they need to be proposed and PR'd separately.

bitemyapp:
This is looking good, just need the changelog and we're good to go. Thank you @Kheldar!

Kheldar:
Changelog line added. I also reverted the whitespace changes - I suspect one of us had rule enabled that did that automatically.

bitemyapp:
@Kheldar thank you very much!

PR #266, opened 2019-10-31
Support es 7

AlexeyRaga:
## Changes

- Upgraded `docker compose` to run ES 7.4.1
- Removed `MappingType`
- Implemented parent/child relationship (it is not perfect, but I have reasons to believe that it has never been perfect in `Bloodhound`)
- Removed support for templated queries because this feature is no longer supported by the ES
- Numerous fixes for annoying incompatibilities and breaking changes

Note that one tests is currently disabled due to a bug that I have found in the ES itself: https://github.com/elastic/elasticsearch/issues/48670
The bug hits rather rare use case, and `bloodhound` is not at fault so I think it shouldn't be a show stopper for moving forward.


AlexeyRaga:
@bitemyapp Sorry, I did not realise that Travis is using a different way of provisioning the ES and that there was an example project.

I have fixed them both and now the build is green, please have another look.

elasticsearch/Dockerfile
```diff
@@ -1,3 +0,0 @@
-FROM docker.elastic.co/elasticsearch/elasticsearch:5.6.0
```
bitemyapp: What's the purpose for deleting stuff behind `elasticsearch/`?

bitemyapp:
@AlexeyRaga I'm interested in supporting this and updating Bloodhound to ES 7. I've asked @JoseD92 to combine his ES 7 update effort with what you've done. So even if we don't merge this specific PR in the end, it's likely the final PR will reflect the good work you've done here.

elasticsearch/Dockerfile
AlexeyRaga: Not needed anymore.
It was used in `docker-compose` only, but now it is possible to just use an official docker image.

AlexeyRaga:
@bitemyapp Where I can see these changes? We are currently maintaining our own fork of `bloodhound` and it would be useful to see what I missed and probably get these changes too until the official release is out.

AlexeyRaga:
With @JoseD92 PR merged, what holds us back from having ES7 support in `master` and on Hackage?

rsoeldner:
@bitemyapp @AlexeyRaga any update on this ?

teto:
that would be awesome, as well as an update on katip-elasticsearch. I can help but would like some confirmation this works first.

bitemyapp:
cc @wraithm

src/Database/Bloodhound/Types.hs
```diff
@@ -542,3 +557,104 @@ instance (FromJSON a) => FromJSON (SearchResult a) where
...
+  parseJSON (String "h") = pure $ TimeUnitHours
+  parseJSON ( String "m") = pure $ TimeUnitMinutes
+  parseJSON (String "s") = pure $ TimeUnitSeconds
+  parseJSON (String "ms") = pure $ TimeUnitMilliseconds
+  parseJSON (String "micros") = pure $ TimeUnitMicroseconds
+  parseJSON (String "nanos") = pure $ TimeUnitNanoseconds
+  parseJSON _          = empty
+
+data SearchTemplate =
+  SearchTemplate {
```
wraithm, line 623: style nitpick: This is a bit different from the other data type definitions in this project. Would you mind changing it?

src/Database/Bloodhound/Types.hs
```diff
@@ -542,3 +557,104 @@ instance (FromJSON a) => FromJSON (SearchResult a) where
...
+instance FromJSON GetTemplateScript where
+  parseJSON (Object v) = do
+    script <- v .:? "script"
+    maybe
+      (GetTemplateScript Nothing Nothing Nothing <$> v .: "_id" <*> v .: "found")
+      (\s -> GetTemplateScript <$>
+        s .:? "lang"    <*>
+        s .:? "source"  <*>
+        s .:? "options" <*>
+        v .: "_id"          <*>
```
wraithm, line 656: Would you mind cleaning this up a bit?

src/Database/Bloodhound/Types.hs
```diff
@@ -542,3 +557,104 @@ instance (FromJSON a) => FromJSON (SearchResult a) where
...
+        v .: "found"
+      )
+      script
+  parseJSON _          = empty
```
wraithm, line 660: Style nit: Please add a newline here

wraithm (commented):
This looks great on first review other than some minor style nitpicks. I'm gonna do some more testing tonight.

teto:
I've tried to use this with katip-elasticsearch and this update breaks it. I wonder where it's best to break the interface bloohound vs katip (I don't know enough of either to have an opinion)

bitemyapp:
cc @MichaelXavier https://github.com/bitemyapp/bloodhound/pull/266#issuecomment-717930767

ivanbakel:
@wraithm I've addressed the style nits you raised.

It seems that the original author is no longer in a position to maintain this particular PR. If there are any other changes that need to be made, I am happy to have a go - but I will need to re-open this under a fork I control. (For example, it seems that the disabled test could be re-enabled).

wraithm:
Great! Thanks. I'll get to this later today.

Makefile
```diff
@@ -98,23 +97,23 @@ upload:
 
 ## Run test environment
 compose-ES5:
```
wraithm, line 99: Minor naming thing, would you mind changing the name of this (and the others) to `compose-ES7`? Maybe it should just simply be called `compose`?

wraithm:
@MichaelXavier Ping, do you have any opinions about katip-elasticsearch and this PR? I've reviewed this, and I'm pretty sure this is good to go.

Makefile
ivanbakel: My personal opinion is that `compose` makes sense because there is nothing obviously connected to the version here.

ivanbakel: This has been addressed on #276, which is a duplicate PR that I control.

PR #268, opened 2020-03-23
Implement Count API

### LINKED ISSUE #267: Support count API
Title says it, docs here: https://www.elastic.co/guide/en/elasticsearch/reference/current/search-count.html.

Looks like an easy enough thing to add, would it be acceptable if I sent a PR?

akshaymankar:
Fixes #267


bitemyapp:
Looks great, thank you!

Issue #273, opened 2020-07-27
Support GHC 8.8


wraithm:
I tried compiling bloodhound with GHC 8.8.3, and I got the following error:

```
    [ 4 of 16] Compiling Database.Bloodhound.Internal.Query

    /path/to/bloodhound/src/Database/Bloodhound/Internal/Query.hs:1626:26: error:
        • Could not deduce (MonadFail m) arising from a use of ‘fail’
          from the context: Monad m
            bound by the type signature for:
                       fieldTagged :: forall (m :: * -> *) a.
                                      Monad m =>
                                      (FieldName -> Object -> m a) -> Object -> m a
            at src/Database/Bloodhound/Internal/Query.hs:1623:1-71
          Possible fix:
            add (MonadFail m) to the context of
              the type signature for:
                fieldTagged :: forall (m :: * -> *) a.
                               Monad m =>
                               (FieldName -> Object -> m a) -> Object -> m a
        • In the expression: fail "Expected object with 1 field-named key"
          In a case alternative:
              _ -> fail "Expected object with 1 field-named key"
          In the expression:
            case HM.toList o of
              [(k, Object o')] -> f (FieldName k) o'
              _ -> fail "Expected object with 1 field-named key"
         |
    1626 |                     _ -> fail "Expected object with 1 field-named key"
         |
```

bitemyapp:
I'd take a patch for this.

PR #274, opened 2020-07-31
Add support for GHC 8.8

akshaymankar:
Fixes https://github.com/bitemyapp/bloodhound/issues/273
This will also require a version bump to quickcheck-arbitrary-template, I created a PR for that here: https://github.com/bitemyapp/quickcheck-arbitrary-template/pull/1. Once that is merged I will update the URL in stack-8.8.yaml and make this PR reviewable.


wraithm:
@akshaymankar There's also https://github.com/plow-technologies/quickcheck-arbitrary-template/pull/9 It'd be great if that got merged and pushed up to hackage first.

akshaymankar:
Given the silence on plow-technologies/quickcheck-arbitrary-template#9, I think it is better to keep bitemyapp/quickcheck-arbitrary-template as the dependency. Maybe it can be changed in a separate PR when the maintainers from plow-technologies respond.

bitemyapp:
Looks like the `UpdatableIndexSetting` isomorphism-ish is on the fritz again. Obnoxious.

akshaymankar:
> Looks like the `UpdatableIndexSetting` isomorphism-ish is on the fritz again. Obnoxious.

The problem is with [`CharFilterDefinitionMapping`](https://github.com/bitemyapp/bloodhound/blob/c5603ff4e120b7b23960373541d8ecfe142f09f3/src/Database/Bloodhound/Internal/Analysis.hs#L63). the test fails whenever the generated `CharFilterDefinition` is constructed with `CharFilterDefinitionMapping` and one of the mappings contains `=>`.  
I think it would be best to represent it as `Map Text (Text, Text)` instead of `Map Text Text`. This would still require a custom instance of `Arbitrary`, as we don't have any way of escaping `=>` (at least it is not documented [here](https://www.elastic.co/guide/en/elasticsearch/reference/current/analysis-mapping-charfilter.html)).

akshaymankar:
@bitemyapp Is it ok if that flakey test is resolved separately?

Daniel-Diaz:
PR https://github.com/plow-technologies/quickcheck-arbitrary-template/pull/9 is now merged and a new version was uploaded to Hackage: http://hackage.haskell.org/package/quickcheck-arbitrary-template-0.2.1.1.

src/Database/Bloodhound/Internal/Query.hs
```diff
@@ -1620,7 +1620,7 @@ functionScoreFunctionsPair (FunctionScoreSingle fn)
 functionScoreFunctionsPair (FunctionScoreMultiple componentFns) =
   ("functions", toJSON componentFns)
 
-fieldTagged :: Monad m => (FieldName -> Object -> m a) -> Object -> m a
+fieldTagged :: (Monad m, MonadFail m)=> (FieldName -> Object -> m a) -> Object -> m a
```
wraithm, line 1623: I tried building this with stack-8.6.yaml, and it failed. I think you'd need to import `MonadFail` for that compiler version.

wraithm:
```haskell
bloodhound          >     Not in scope: type constructor or class ‘MonadFail’
bloodhound          >     Perhaps you meant ‘MonadFix’ (imported from Bloodhound.Import)
bloodhound          >      |
bloodhound          > 1623 | fieldTagged :: (Monad m, MonadFail m)=> (FieldName -> Object -> m a) -> Object -> m a
bloodhound          >      |```

akshaymankar: Thanks for finding this. I also went through older versions and adjusted imports accordingly. I also bumped LTS versions for them, only GHC 8.0 had a major LTS version bumped, it compiles, so hopefully, it shouldn't be a problem.

wraithm (approved):

PR #281, opened 2021-08-31
Point in time (PIT) API support

mirokuratczyk:
Adds support for https://www.elastic.co/guide/en/elasticsearch/reference/current/point-in-time-api.html, which is now recommended over the scroll API for deep pagination.

I considered introducing a new exception type, or returning an error, instead of using `error` in `pitSearch` and `pitAccumulator`, but the cases they handle _should_ be impossible.

In the future it would be nice to introduce [advanceScroll](https://github.com/bitemyapp/bloodhound/blob/4775ebb759fe1b7cb5f880e4a41044b2363d98af/src/Database/Bloodhound/Client.hs#L1169) type functionality for the PIT API. This will require [passing some state around](https://github.com/mirokuratczyk/bloodhound/blob/36c682c63cb0268c3b9711403be7fbd608235666/src/Database/Bloodhound/Client.hs#L1241-L1243) between calls so I left it out for now, because this change should be sufficient for using the API.

We've been relying on this change in our fork so it would be great to get it merged so we can point to upstream again.

### DIFF
--- a/bloodhound.cabal
+++ b/bloodhound.cabal
@@ -44,6 +44,7 @@ library
       Bloodhound.Import
       Database.Bloodhound.Common.Script
       Database.Bloodhound.Internal.Count
+      Database.Bloodhound.Internal.PointInTime
       Paths_bloodhound
   hs-source-dirs:
       src
--- a/src/Database/Bloodhound/Client.hs
+++ b/src/Database/Bloodhound/Client.hs
@@ -64,6 +64,9 @@ module Database.Bloodhound.Client
        , getInitialScroll
        , getInitialSortedScroll
        , advanceScroll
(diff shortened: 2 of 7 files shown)


blackheaven:
Hello @mirokuratczyk 

If you find some time to rebase your PR this week, I'll be glad to merge it, otherwise, I'll handle it myself this week-end.

Regards.

blackheaven (approved):

blackheaven:
Thanks, I'll run the tests and merge it tomorrow.

blackheaven:
Thanks for your contribution

mirokuratczyk:
Thanks for reviewing and merging this!

PR #285, opened 2022-03-31
Update aeson min-bound to >=2.0

TristanCacqueray:
This change prevents compilation error when building with aeson-1:

  src/Bloodhound/Import.hs:28:1: error:
      Could not find module ‘Data.Aeson.Key’


blackheaven (approved):

blackheaven:
Sorry, I missed it, thanks!

PR #291, opened 2023-08-08
_type is not always sent by ES

robwithhair:
It appears that ES used to send _type but to some requests no longer does.

I presume this should become a Maybe type.  Unfotunately I can't run the tests locally because of missing stack.yaml, though I did follow the instructions in Readme.


blackheaven:
Sorry, I screwed the squash.

I cherry-picked it as 3866f85 thanks a lot.

PR #296, opened 2024-10-17
chore: Let shell.nix depend on nix flake

supersven:
It's usually better to have one source of truth (regarding package sources/versions) than two. So, use the compatibility mixin for those who want to keep using shell.nix.

GHC 9.2 is not supported anymore ("end of life".) Thus, we don't need overrides for it.

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -1,5 +1,19 @@
 {
   "nodes": {
+    "flake-compat": {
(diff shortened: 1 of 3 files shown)


blackheaven (approved):

blackheaven:
Thanks for your contribution!

PR #297, opened 2024-10-17
feature: Add docker-compose to flake / devShell

supersven:
`dockerd` relies on superuser/system access and thus should be provided by the operating system. However, `docker-compose` simply uses `dockerd` and thus can be part of the flake's dependencies for the `devShell`.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -44,13 +44,14 @@
               };
             in
             pkgs.mkShell {
-              buildInputs = [pkgs.docker-compose] ++ (with haskellPackages; [
+              buildInputs = with haskellPackages; [
+                pkgs.docker-compose
                 haskell-language-server
                 ghcid
                 cabal-install
                 scripts
                 ormolu
-              ]);
+              ];
               inputsFrom = [
                 (enableTests self.defaultPackage.${system}).env
               ];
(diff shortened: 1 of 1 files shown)


flake.nix
```diff
@@ -44,13 +44,13 @@
               };
             in
             pkgs.mkShell {
-              buildInputs = with haskellPackages; [
+              buildInputs = [pkgs.docker-compose] ++ (with haskellPackages; [
                 haskell-language-server
                 ghcid
```
blackheaven, line 49:
```suggestion
              buildInputs = with haskellPackages; [
                pkgs.docker-compose
                haskell-language-server
                ghcid
```

flake.nix
```diff
@@ -44,13 +44,13 @@
...
                 cabal-install
                 scripts
                 ormolu
-              ];
+              ]);
```
blackheaven, line 53:
```suggestion
              ];
```

blackheaven (changes requested):

flake.nix
supersven: That should be fine as well :+1:

blackheaven (approved):

blackheaven:
Thanks for your contributions.

Issue #295, opened 2024-10-16
`parseEsResponse` cannot parse version conflict error response (OpenSearch 1.3)


supersven:
Bloodhound version: 76a27a44223e8c24ec2c6a13504ea671887f2672
OpenSearch version: 1.3.19

Error:
```
EsProtocolException {
   esProtoExMessage = "Original error was: Non-200 status code Error parse failure was: Error in $: key \"error\" not found"
   , esProtoExResponse = "{\"took\":9,\"timed_out\":false,\"total\":3,\"updated\":1,\"deleted\":0,\"batches\":1,\"version_conflicts\":2,\"noops\":0,\"retries\":{\"bulk\":0,\"search\":0},\"throttled_millis\":0,\"requests_per_second\":-1.0,\"throttled_until_millis\":0,\"failures\":[{\"index\":\"directory_test\",\"type\":\"_doc\",\"id\":\"9fda4188-2afd-490d-8796-e023df61a4e9\",\"cause\":{\"type\":\"version_conflict_engine_exception\",\"reason\":\"[9fda4188-2afd-490d-8796-e023df61a4e9]: version conflict, required seqNo [11], primary term [1]. current document has seqNo [16] and primary term [1]\",\"index\":\"directory_test\",\"shard\":\"0\",\"index_uuid\":\"Y3RpVY_DQEW9ULn8oGulrg\"},\"status\":409},{\"index\":\"directory_test\",\"type\":\"_doc\",\"id\":\"d70b631d-966a-4951-a94c-35ddc210f28a\",\"cause\":{\"type\":\"version_conflict_engine_exception\",\"reason\":\"[d70b631d-966a-4951-a94c-35ddc210f28a]: version conflict, required seqNo [13], primary term [1]. current document has seqNo [15] and primary term [1]\",\"index\":\"directory_test\",\"shard\":\"0\",\"index_uuid\":\"Y3RpVY_DQEW9ULn8oGulrg\"},\"status\":409}]}"
}
```

supersven:
To me it looks like the `FromJSON EsError` instance does not fit to the response.

```haskell
instance FromJSON EsError where
  parseJSON (Object v) =
    EsError
      <$> v
        .:? "status"
      <*> (v .: "error" <|> (v .: "error" >>= (.: "reason")))
  parseJSON _ = empty
```

The implementation expects an `error` or `error.reason` field, while in the response there's an array (`failures`)  with `reasons`.

[...]

PR #301, opened 2024-10-18
fix: Parsing EsError for OpenSearch 1.3

supersven:
The error format changed.

The solution isn't ideal as it drops all but the first failure. However, it's better than having to deal with `EsProtocolException`s.

The this would fix https://github.com/bitemyapp/bloodhound/issues/295


blackheaven (approved):
It's not what I have in mind, but that's a first step, thanks.

PR #307, opened 2024-11-05
fix: Delete confusing assertion in test

supersven:
The assertion implied that a version conflict can lead to a Right value (maybe for either ElasticSearch or OpenSearch.) However, the parsing logic would force this to be a Left value.

### DIFF
--- a/tests/Test/DocumentsSpec.hs
+++ b/tests/Test/DocumentsSpec.hs
@@ -35,7 +35,7 @@ spec =
         liftIO $ isCreated res `shouldBe` True
         insertedConflict <- tryEsError $ insertData' cfg
         case insertedConflict of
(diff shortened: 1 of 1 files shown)


tests/Test/DocumentsSpec.hs
```diff
@@ -36,6 +36,7 @@ spec =
         insertedConflict <- tryEsError $ insertData' cfg
         case insertedConflict of
           Right (res', _) -> liftIO $ isVersionConflict res' `shouldBe` True
+          Right _ -> liftIO $ expectationFailure "This should never not happen."
```
blackheaven, line 39: Isn't it redundant?

supersven: That's a *too much Git magic to create small PRs*-mistake. Sorry for the noise - I'm fixing it.

